PR #13, opened 2017-12-17
RTLSDR Gain Fixes & Other updates.

darksidelemm:
- Fixed issues with rtlsdr_gain setting. Use a setting of -1 to get automatic gain control, otherwise specify the desired gain as a value in dB.
- Added the plot_rtl_power.py util in utils/ to assist with debugging these sorts of issues.
- Added a dwell_time parameter, which allows adjusting the time rs_detect sits on a frequency and tries to detect a sonde. Allows for speeding up scanning quite considerably. Needs more testing to determine a suitable default value.
- Added detection of OSX OS, which removes the '-k 30' option from the rtl_power call, as this isn't supported under the macports-provided timeout command. May need to look at a better way of handling this.
- Made the -f option quit if no sonde is detected on the supplied frequency.

### DIFF
--- a/auto_rx/config_reader.py
+++ b/auto_rx/config_reader.py
@@ -12,7 +12,7 @@ def read_auto_rx_config(filename):
 	# Configuration Defaults:
 	auto_rx_config = {
 		'rtlsdr_ppm'	:	0,
-		'rtlsdr_gain'	:	0,
+		'rtlsdr_gain'	:	-1,
 		'rtlsdr_bias'	: False,
 		'search_attempts':	5,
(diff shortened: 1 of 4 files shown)


auto_rx/config_reader.py
```diff
@@ -47,8 +48,8 @@ def read_auto_rx_config(filename):
 		config = ConfigParser.RawConfigParser()
```
TheSkorm: Should we change this to `config = ConfigParser.RawConfigParser(auto_rx_config)` so that it uses the defaults if a config option is missing from the config file?

PR #68, opened 2018-06-01
auto_rx modular restructure.

darksidelemm:
Big re-structure to allow easier upgrades in the future.
Main improvements:
- Python 3 compatibility!
- Support for using multiple RTLSDRs for tracking of multiple sonde launches. (With some caveats)
- Code moved to a somewhat more modular architecture.


TheSkorm:
Functionally it looks ok, however I am getting
```
ERROR:lsusb parsing error: unexpected indentation: "   Port 8: 0000.0100 power"
```

which might confuse people

auto_rx/station.cfg.example
```diff
@@ -151,19 +160,46 @@ max_altitude = 50000
...
+# Only accept a payload ID as valid until it has been seen N times.
+# This helps avoid corrupted callsigns getting onto the map.
+payload_id_valid = 5
+# Paths to the rtl_fm and rtl_power utilities. If these are on your system path, then you don't need to change these.
+sdr_fm_path = rtl_fm
+sdr_power_path = rtl_power
+
 
-# Rotator Settings
+# Rotator Settings - Currently not implemented.
```
TheSkorm: should we have rotator before we merge - does anyone use it?

darksidelemm: Not sure if there are any users of it at the moment. Don't think so. 
I'll get it re-added once I have my new Az/El system up, and can actually test the code.

PR #75, opened 2018-07-04
Add email notification function

fsphil:
Sends an email when a new sonde is detected, with some basic details and a link to habhub.


darksidelemm:
Could you please move the new config reader lines to a try/except block, so if users have not updated their configuration file they don't get an immediate error. 

See here for an example: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/config.py#L190

We'll then merge the new configuration file fields in at a later date.

fsphil:
Updated now. It looks to see if the enable flag exists first, then silently disables emails if it isn't there.

PR #84, opened 2018-08-01
Add az/el/range calculation, and LOS bearing on map.

darksidelemm:
- Add the ability to show azimuth/elevation/range data, if the station lat/lon is provided.
- Add a line-of-sight path line, like on tracker.habhub.org
- Add age field (this is a bit hacky, I have to update the entire table every second for this to work.. doesn't appear to add much overhead)
- Reverse ordering on log table, so newest log entries appear at top.
- Add "Hide map" button, and move hide map/scan plot buttons *above* their respective divs.

### DIFF
--- a/auto_rx/autorx/templates/index.html
+++ b/auto_rx/autorx/templates/index.html
@@ -72,9 +72,9 @@
             socket.on('log_event', function(msg) {
                 // On receipt, add to the log table.
                 var log_entry = "<tr><td class='col-2'>"+msg.timestamp+"</td><td class='col-1'>"+msg.level+"</td><td class='col-9'>"+msg.msg+"</td></tr>";
-                $('#log_data > tbody').append(log_entry);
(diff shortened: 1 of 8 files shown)


auto_rx/autorx/templates/index.html
```diff
@@ -338,6 +396,12 @@
             }, sonde_follow_timeout);            
 
 
+            // Update telemetry table every 500ms
```
TheSkorm: This comment seems wrong. Seems to be every second.

Do we do this because we don't want the table updating so often that the user can't read the data in it if it's constantly changing?

darksidelemm:
Was hitting some issues where the web client would hang after a long time of no activity, so have slowed down the table update rate to every 5 seconds instead of every second.
Have also added a check for the RS binaries (rs_detect, etc).

Please update and test.

PR #96, opened 2018-12-20
WIP: Viproz M10 support

TheSkorm:
Creating a pull request to track changes required


Viproz:
This PR is fully working, the detection of M10 is working it gets the right frequency, decoding working too.

I am going to keep on making improvement to the decoding algorithm (for example the frame detector is lagging behind at the moment) but it is working pretty well in the current state.

The ID used for APRS is what I call the dxl ID because dxl is supporting M10 with trimble GPS (named ptu in the code) and using this ID. On any other places the ID is going to be the real ID which contains the time of fabrication and the serial number( see here for more details https://github.com/rs1729/RS/issues/7).

TheSkorm:
Thanks heaps @Viproz 
I've merged this into our testing branch and we'll test this branch on some of our stations to ensure there are no regressions before merging in to master.

Thanks for the quality code and testing you've done so far, we greatly appreciated it!

Issue #98, opened 2018-12-25
GEvent Messages (was: HTTP output in log)


SA7BNT:
Merry Christmas @all 

im getting this output in my log file. Is that normal or is it a issue on my side??
![bildschirmfoto 2018-12-25 um 07 55 41](https://user-images.githubusercontent.com/35367117/50415377-ed9a6000-081a-11e9-8b9d-4b3b1b6f60e7.png)

darksidelemm:
Hmm, you definitely shouldn't be seeing any output from the web-server, we should be inhibit messages from it (apart from errors) with these few lines: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/auto_rx.py#L397

What do you see right when auto_rx starts up? It should give some output regarding the flask webserver that might shed some more light on the situation.

SA7BNT:
Hey darksidelemm, 
here is the output.
![bildschirmfoto 2018-12-25 um 11 04 34](https://user-images.githubusercontent.com/35367117/50420158-0c0d5500-0835-11e9-98f0-aac27f147724.png)

And this is coming up when im open the browser
![bildschirmfoto 2018-12-25 um 11 04 51](https://user-images.githubusercontent.com/35367117/50420186-2e9f6e00-0835-11e9-9479-e9a99aa38bfc.png)

Viproz:
I can confirm the issue : https://pastebin.com/tj2ia39X
Theses logs were obtained after a reload of the webpage

darksidelemm:
OK, so Viproz has confirmed the messages are coming from geventwebsocket, which suggests something has pulled in gevent as a dependency. I'm not sure what at this point.

I've only briefly tried running the web interface using gevent as the webserver without much success. That it's working at all is pretty interesting. 

Noting the heap of crap in your log files, could you try adding the following after the block of 'logging.getLogger' lines at line 403 of auto_rx.py :

darksidelemm:
It's also possible to force the use of werkzeug as the server by instantiating the SocketIO object with the async_mode argument of 'threading'.

i.e. change the line https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/web.py#L40
to:

socketio = SocketIO(app, async_mode='threading')

This is not an ideal solution - gevent is probably a better server to use, though it would be a suitable workaround. The werkzeug server is sufficient for the very small number of concurrent users expected to access the auto_rx web interface.

[...]

PR #99, opened 2018-12-26
Use werkzeug by default, suppress any gevent messages.

darksidelemm:
Small pull request to temporarily fix https://github.com/projecthorus/radiosonde_auto_rx/issues/98

Really needs testing by someone that has gevent installed to ensure the forcing of werkzeug works.


Viproz:
I can confirm the fix is working, I had the issue before and ctrl+c wasn't working, after the fix no more logs and ctrl+c is working. Here are the logs show with and without verbose (-v) : 
https://pastebin.com/qJvzy5y7

Steps taken : 
- New clean clone
- Launch python script
- Reload webpage
- Ctrl+c terminal

Issue #105, opened 2019-01-27
Integration with Balena.io - fleet management RPi


SloMusti:
To simplify the deployment of the receiver and enable a more straightforward deployment, I have set-up a prototype docker deployment that can be used with http://balena.io system to mange a fleet of devices.

To do so, there were minimal changes required and it would be great to add support to the master that supports this. The changes made: https://github.com/SloMusti/radiosonde_auto_rx/commits/master
This in particular is just changing the default log path as `./log/` is not available when running the system in Balena, but must be placed on the data partition, thus `/data/log/` works. I suggest the log path should be configurable in settings.

The other required change is to make sure the config file is also on path `/data/` where it can be user modified.

To get this running on Balena.io system and deploy a fleet, I have published the code here: https://github.com/SloMusti/multicontainer-lora-radiosonde with the following key features:
 * Adds Balena.io support
 * Adds radiosonde_auto_rx support
 * Makes radiosonde web interface available via public URL option of balena
 * Adds TheThingsNetwork LoraWAN gateway support in parallel on the same device if you have that receiver installed

I am open to suggestions how to best implement the log support and can prepare the pull request then.

darksidelemm:
Hi! 
Thanks for your work on this, this is amazing! Easy deployment of auto_rx has always been one of the things we needed to improve, and it looks like you're pretty much got the problem solved!

I have no issues with the log file directory being a configuration file parameter, in fact thats something I have been thinking about for a while. This would require changes to autorx/config.py to read in the path, and the other necessary changes where log files are written (in auto_rx.py and also in autorx/logger.py)

As for the configuration file, this can be specified with the --config option as it looks like you've found already. Is this sufficient?

Cheers,
Mark

SloMusti:
Sure `--config` is suitable, best would probably be to have also the  `--log` option that overrides the log folder as that would probably keep the code backwards-compatible for existing users.

Also, it is quite easy to add the following to the code: https://github.com/balena-io-projects/balena-dash so the receiver can have a nice display showing current results.

darksidelemm:
OK, breaking out out the log directory setting to --log is a good solution.

What changes did you have in mind for the dash support? I'd rather not add too much solution-specific stuff to the main auto_rx code if you get my meaning. Rather, i'd prefer to add suitable interfaces to enable other displays to be written. We have some of this already with the UDP output option, which is used by chasemapper.

SloMusti:
[...]

PR #106, opened 2019-01-27
Implemented command line argument log folder

SloMusti:
As discussed in  #105 I have implemented the option to configure the log folder in the command line interface.

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -384,7 +384,7 @@ def main():
 
 
     # Configure logging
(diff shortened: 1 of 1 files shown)


darksidelemm:
Could you please set the default configuration file location back to "station.cfg" ?
Here: https://github.com/projecthorus/radiosonde_auto_rx/pull/106/files#diff-60891b14d50d29843216596ac4c2a329R368

Otherwise this will break current installs.

SloMusti:
Well spotted, I have restored the correct default.

PR #129, opened 2019-03-03
GPS satellites number and battery level

Viproz:
This is a PR to add the satellites number and the battery level to M10 Trimbles (or new GPS chip but same modulation), I added them as "satellites" and "battery" in the json.
I'm open for critiques for the names but I think we should uniformize theses names as they are both useful for RS hunting, the GPS number can tell us how precise the position emitted by the RS is and the battery level tells us if everything is OK and the RS will continue emit for a while.

We could send theses infos in the comments, maybe not each frame but once a while.

### DIFF
--- a/m10/M10GtopParser.cpp
+++ b/m10/M10GtopParser.cpp
@@ -192,6 +192,16 @@ std::string M10GtopParser::getSerialNumber() {
         sn_bytes[i] = byte;
     }
 
+    /*
+     * The serial number is in the form M10-A-BCC-D-EEEE
+     * - A is the frame type, T for Trimble, the original GPS used for this modulation
+     * G for Gtop GPS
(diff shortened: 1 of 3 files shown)


darksidelemm:
On the RS41 decoder the number of sats field is currently called 'sats'. 
I was sure I had added in some code into the filter function to discard positions with <4 sats: https://github.com/projecthorus/radiosonde_auto_rx/blob/testing/auto_rx/auto_rx.py#L314

.. but I think I must have reverted the change accidentally at some point. I'll add it back into the testing branch. 

We need to be very careful changing fields in the uploaded telemetry, as we would need a large mass of users to update at the same time to avoid weirdness on the online trackers, however changes to show it on the local web interface are easy enough. Changes to only the comment field are probably fairly safe though.

PR #210, opened 2019-09-09
Iterate over a copy of the dictionary

snh:
Without this, Python 3.7.3 is throwing the following error after a pop:

```
RuntimeError: dictionary changed size during iteration
```

### DIFF
--- a/auto_rx/autorx/logger.py
+++ b/auto_rx/autorx/logger.py
@@ -201,7 +201,7 @@ class TelemetryLogger(object):
 
         _now = time.time()
 
-        for _id in self.open_logs.keys():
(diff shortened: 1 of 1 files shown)


darksidelemm:
Now I'm wondering where else I've got this issue...
I will admit that auto_rx hasn't been tested thoroughly under Python 3.7.

snh:
@darksidelemm That is the only one I've tripped over, but now you mention it I do see other places in the codebase that I suspect would experience the same issue.

Another Pull Request will be incoming shortly 😄.

darksidelemm:
Could you make the PR to the testing branch please? I'm currently in the process of adding support for a new sonde :-)

PR #211, opened 2019-09-09
Iterate over a copy of the dictionary (2nd instalment)

snh:
Followup to https://github.com/projecthorus/radiosonde_auto_rx/pull/210, with the aim of proactively dealing with other occurances where a `pop` will cause a dictionary to be modified while being iterated over.

As far as I know I hadn't experienced any issues with any of these yet, but it may just be I haven't exercised the relevant code-paths in my simple usage so far. With that in mind, these changes should be assumed to be untested.

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -325,7 +325,7 @@ def handle_scan_results():
 def clean_task_list():
     """ Check the task list to see if any tasks have stopped running. If so, release the associated SDR """
 
-    for _key in autorx.task_list.keys():
+    for _key in autorx.task_list.copy().keys():
         # Attempt to get the state of the task
         try:
             _running = autorx.task_list[_key]['task'].running()
@@ -358,7 +358,7 @@ def clean_task_list():
             flask_emit_event('task_event')
 
     # Clean out the temporary block list of old entries.
-    for _freq in temporary_block_list.keys():
+    for _freq in temporary_block_list.copy().keys():
         if temporary_block_list[_freq] < (time.time() - config['temporary_block_time']*60):
             temporary_block_list.pop(_freq)
             logging.info("Task Manager - Removed %.3f MHz from temporary block list." % (_freq/1e6))
--- a/auto_rx/autorx/scan.py
+++ b/auto_rx/autorx/scan.py
@@ -657,7 +657,7 @@ class SondeScanner(object):
             
             # Remove any frequencies in the temporary block list
             self.temporary_block_list_lock.acquire()
-            for _frequency in self.temporary_block_list.keys():
+            for _frequency in self.temporary_block_list.copy().keys():
                 # Check the time the block was added.
                 if self.temporary_block_list[_frequency] > (time.time()-self.temporary_block_time*60):
                     # We should still be blocking this frequency, so remove any peaks with this frequency.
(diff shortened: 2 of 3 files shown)


darksidelemm:
This one is probably not required: https://github.com/projecthorus/radiosonde_auto_rx/pull/211/files#diff-0291e9635ed258874ca6302bbdab2425L125
The dictionary in question was already copied in the previous line.

Will check over the others...

darksidelemm:
The access of autorx.task_list in clean_task_list() should be safe - this and handle_scan_results() are the only two functions that add or remove elements from autorx.task_list, and they do it in the main loop (i.e. here https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/auto_rx.py#L721 )

Having a look at other places where autorx.task_list is *read* however, the obvious one that could break is the /get_task_list flask endpoint here: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/web.py#L78 which iterates over the task list - it's could be possible for the task list to be changed during that loop. However if it is changed, and that function call throws an exception, it's not going to break the whole of auto_rx - flask will just return a HTTP 500 error...

The temporary_block_list access in scan.py should be safe, as all accesses to the block list are handled using a Lock. This was done because entries can be added to this list from the main thread, via this function: https://github.com/projecthorus/radiosonde_auto_rx/blob/bdbfbb03f019ba6710086602ae582c2b5436ad91/auto_rx/autorx/scan.py#L789  (which again, acquires the lock before accessing the list)

The access of temporary_block_list in clean_task_list() is probably OK too, as that's the only function that adds/removes entries from that list. The list is passed into a Scanner instance when initialised, but the list is copied, not used by reference. (see here: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/scan.py#L445 )

snh:
> This one is probably not required: https://github.com/projecthorus/radiosonde_auto_rx/pull/211/files#diff-0291e9635ed258874ca6302bbdab2425L125
> The dictionary in question was already copied in the previous line.

Indeed, but we appear to be modifying the copy, so will still have this problem afaik, as we are modifying the dictionary we are itirating over.

With regards to the others, I don't think it is just a case of whether or not something else is accessing these similtanously, but rather that we are modifying the object we are itirating over, so even in a single threaded context this is still an issue afaik.

I just happened to trigger the `clean_task_list` one just then:

```
Traceback (most recent call last):
  File "/opt/auto_rx/auto_rx.py", line 741, in <module>
    main()
  File "/opt/auto_rx/auto_rx.py", line 721, in main
    clean_task_list()
  File "/opt/auto_rx/auto_rx.py", line 328, in clean_task_list
    for _key in autorx.task_list.keys():
RuntimeError: dictionary changed size during iteration
```

snh:
Just realised with the one in `web.py` we are performing the pop within a sub-dictionary, rather than the `_temp_store` dictionary which we are iterating over, so the `.copy()` shouldn't be necesary in this one afterall.

darksidelemm:
Ahhhhh. Now I understand what you mean. Yes, of course that makes sense with modifying the dictionary within the loop..  and I haven't hit this issue before because I haven't been running auto_rx for long periods under Python 3.7... I'd better start using Python 3.7 as my usual development python from now on. Thanks for the help :-)

PR #227, opened 2019-10-04
Update gitignore

TheSkorm:


darksidelemm:
You missed demod/mod/m10mod, which we'll be using at some point soon.

PR #253, opened 2019-11-14
Moved log_power.csv to /tmp/

Luceoria:
This should reduce SD card I/O a bit. If you find this pointless, please disregard.


darksidelemm:
Not a terrible idea, but the change you need to make is here: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/scan.py#L630

darksidelemm:
Having some laptop issues at the moment, will do some testing on a RPi when I get a chance.

Can you re-point this PR at the testing branch? This is where I prefer to stage updates for testing prior to release.

Luceoria:
Sure, will do.

PR #255, opened 2019-11-18
Moved log_power.csv to /tmp/

Luceoria:
This should reduce SD card I/O a bit.

### DIFF
--- a/auto_rx/autorx/scan.py
+++ b/auto_rx/autorx/scan.py
@@ -627,7 +627,7 @@ class SondeScanner(object):
             run_rtl_power(self.min_freq*1e6,
                 self.max_freq*1e6,
                 self.search_step,
-                filename="log_power_%s.csv" % self.device_idx,
+                filename="/tmp/log_power_%s.csv" % self.device_idx,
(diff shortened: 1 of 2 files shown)


darksidelemm:
Sorry to be a pain, I've had some discussions as to the best way to deal with this. 
The 'correct' way, would be to use the tempfile.gettempdir() function to find the appropriate directory for temporary files. See here: https://docs.python.org/3.8/library/tempfile.html#tempfile.gettempdir

This function returns a directory name, e.g.
On Linux:
```
>>> from tempfile import gettempdir
>>> gettempdir()
'/tmp'
```

On OSX:
```
>>> from tempfile import gettempdir
>>> gettempdir()
'/var/folders/g5/f1sl_t4x2yn65bnb66pkwssm0000gn/T'
```

On Windows:
```
>>> from tempfile import gettempdir
>>> gettempdir()
'c:\\users\\vk5qi\\appdata\\local\\temp'
```

You would need to then join this directory name with the filename using os.path.join().

This way the correct temp file directory for each OS is used, and auto_rx remains as cross-platform as possible.

PR #272, opened 2020-04-16
The host spacenear.us no longer resolves, use legacy-snus.habhub.org …

fsphil:
…instead.


darksidelemm:
This function shouldn't actually matter any more, as we no longer create flight documents, and so don't have to check if a sonde has been observed on the map. I'm happy to merge it anyway.

PR #293, opened 2020-08-01
Fix log table

pecord:
When trying to view the log table from an iphone or rpi + touch screen the columns would get smashed.

Before - 
Screen Shot 2020-07-31 at 7 45 29 PM

I made these changes and it renders better - 
Screen Shot 2020-07-31 at 8 58 26 PM


darksidelemm:
With this PR the log area now just gets continually bigger and bigger - we've lost the scrollbars that appear after a few entries (~8 entries or so on my system).

Can you try and bring these back please? Some people leave the auto_rx interface open in a browser window for long periods.

darksidelemm:
Also an easy way to test what happens with many log entries is to run auto_rx.py with the -v option. This way you see all debug messages, which fill up the log quickly.

pecord:
Thanks for the feedback, I’ll check it out tonight

darksidelemm:
I think this one has been overtaken by other updates..

PR #297, opened 2020-09-21
simplify script

hsmade:
little trick to change to change to the directory that the script exists in.


darksidelemm:
Didn't even know that was possible, so thanks 👍 

I've retargeted this to the testing branch, which will get merged into master 'soon enough' (once I fix a few other things)/

PR #304, opened 2020-10-06
Add KML capability

drid:
Adds 2 routes that provide KML files.
 **/rs.kml** provides a network KML element with autorefresh
**/rs_feed.kml** provides placemarks populated by tracked radiosondes

Decoder was extended with a methode to pull telemetry data. I am not sure wether this is the best way to get the data but followed this approach since KML file updates are polled from client

There where several white spaces that were cleaned up by my editor. They are in a commit of their own so it can be reverted.

### DIFF
--- a/auto_rx/autorx/decode.py
+++ b/auto_rx/autorx/decode.py
@@ -175,7 +175,6 @@ class SondeDecoder(object):
         self.rs41_drift_tweak = rs41_drift_tweak
         self.experimental_decoder = experimental_decoder
         self.imet_location = imet_location
-        self.telemetry = {}
 
         # iMet ID store. We latch in the first iMet ID we calculate, to avoid issues with iMet-1-RS units
         # which don't necessarily have a consistent packet count to time increment ratio.
@@ -1162,11 +1161,11 @@ class SondeDecoder(object):
                             _exporter(_telemetry)
                         except Exception as e:
                             self.log_error("Exporter Error %s" % str(e))
-            self.telemetry = _telemetry
+
             return _telem_ok
 
     def log_debug(self, line):
-        """ Helper function to log a debug message with a descriptive heading.
+        """ Helper function to log a debug message with a descriptive heading. 
         Args:
             line (str): Message to be logged.
         """
@@ -1203,16 +1202,13 @@ class SondeDecoder(object):
             self.decoder.join()
 
     def running(self):
-        """ Check if the decoder subprocess is running.
+        """ Check if the decoder subprocess is running. 
 
         Returns:
             bool: True if the decoder subprocess is running.
         """
         return self.decoder_running
 
-    def getTelemetry(self):
-        """ Return telemetry data """
-        return self.telemetry
 
 if __name__ == "__main__":
     # Test script.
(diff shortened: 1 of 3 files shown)


darksidelemm:
The way you're getting the current position / telemetry is fine (probably the easiest way), but it would also be good to get the historical track too, even if only position. It might be possible to merge in the data used by the web interface somehow?

drid:
Yes, I will try to.
Is there a way for feeding data for testing without involving the radio part?

darksidelemm:
Unfortunately not that easy... I do have some scripts squirreled away somewhere which emulate the output of rtl_fm, but i'm not sure where they are. 
I usually just turn on a captured radiosonde!

darksidelemm:
Switching to including a requirements.txt has been on the todo list for a while :-P
(As is re-organising the entire repo to be a proper python package... but thats a lot of work)

darksidelemm:
A few questions:
- What does flask.request.host_url return? Is it just http://0.0.0.0/ or http://localhost/, or does it work out the systems IP? (e.g. will accessing the dynamic KML work from another machine)
- Can you look into adding historical track data too? The track data is available in flask_telemetry_store: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/web.py#L45  It should be easy enough to create a path in the KML with each of the tracks. It's indexed by sonde ID.

drid:
It will return whatever was in the request so 
```
curl http://rx.station:5001/rs.kml
<?xml version="1.0" encoding="UTF-8"?>
<kml xmlns="http://www.opengis.net/kml/2.2" xmlns:gx="http://www.google.com/kml/ext/2.2">
    <Document id="11">
        <NetworkLink id="12">
            <name>Radiosonde AutoRX</name>
            <Link id="13">
                <href>http://rx.station:5001/rs_feed.kml</href>
                <refreshMode>onInterval</refreshMode>
                <refreshInterval>10</refreshInterval>
            </Link>
        </NetworkLink>
    </Document>
</kml>
```

If it is requested by ip, then it will reply with an ip url

Looking into track data

drid:
Added history as linestring.
Tried with track but unless view is continuously centred on object, it will not display last position.
Track can also contain telemetry parameters but these need to be added to every coordinate entry in track object so I'll revisit this at a later time
Here is a sample view
image

darksidelemm:
Looking good! I'll have a play with this on the weekend.

darksidelemm:
I've manually merged this into the testing branch, and I also ran black across all the python source to deal with formatting and style compliance issues. 
I also added a link to the Live KML to the web interface.

I'll give it a try on tonight's launch, but testing with a sonde on the bench it seems to be working fine! Thanks for all your work on this!

drid:
Thinking of adding a line of sight line between HAB and ground station

darksidelemm:
If you think it's not going to look too cluttered (or if you can turn it off), go for it!

darksidelemm:
Can you rebase this PR against the current testing branch? I made a few small changes to the KML interface to make telemetry be fixed point, etc.

drid:
Yes I am doing that now

drid:
Looks ok

PR #317, opened 2020-11-24
Remove timeout from the start script

marcelrv:
After spending some hours trying to find the cause of auto_rx stopping each time before a balloon can be received I finally found that in the start script the timeout defined in the station.cfg is overwritten. Hopefully this PR prevents the next person to waste this time...

### DIFF
--- a/auto_rx/auto_rx.sh
+++ b/auto_rx/auto_rx.sh
@@ -15,5 +15,5 @@ cd /home/pi/radiosonde_auto_rx/auto_rx/
 # Clean up old files
 rm log_power*.csv
 
-# Start auto_rx process with a 3 hour timeout.
-python auto_rx.py -t 180
+# Start auto_rx process
+python auto_rx.py


darksidelemm:
There's a reason that time out is there.

Some people start up auto_rx via a cronjob (see here: https://github.com/projecthorus/radiosonde_auto_rx/wiki#352-option-2---operation-via-crontab-backup-option-if-systemd-not-available---not-recommended ) so it's only running at times of the day when sondes are flying in their area. 

In this case, the timeout is required to ensure auto_rx shuts down after some time.

If using systemd, you shouldn't be using the script at all, and instead calling the python script directly (as per https://github.com/projecthorus/radiosonde_auto_rx/wiki#351-option-1---operation-as-a-systemd-service-recommended )

It's sounding like i should just remove the script entirely to avoid confusion!

marcelrv:
Yes, the reason is clear. The confusing part is that there is the setting in the station.cfg for 180min as well as in the shell file.
As you are already overwriting it in the systemd file to 0 (which makes sense, as you woulnt want it to stop), the one in the .sh is bit double.

Anyway, removing the script completely also fixes the confusion as I would have just run the python file.

darksidelemm:
Actually, looking into this more, the script is required to ensure that auto_rx.py is run from the correct directory, so it needs to stay.

Also worth pointing out that the rx_timeout value in station.cfg is for the individual decoder, e.g. timeout before it goes back into scan mode. I'll clarify this in the config file.

The -t argument is a 'global' exit timer, so auto_rx will close after that time regardless of if anything is being decoded. As it turns out the default for that is 0, so the `-t 0` in the systemd service example is redundant.

PR #323, opened 2020-12-10
APRS decode

YO3IBZ:
Hi,

can you make it to decode also APRS?

Thanks.
Aurel.


darksidelemm:
I'm not sure why you raised this as a pull request, but no, I have no intention of making auto_rx decode APRS. There are already many other well functioning decoders for APRS reception (e.g. direwolf).

YO3IBZ:
if you make it to decode APRS and WSPR, we can use it for ARHAB balloons :)
Thanks.

darksidelemm:
IMO it doesn't really fit well in the auto_rx project. 
If you want to decode ARHAB balloons during 'off' periods, then you're better off using some other decoder, and setting up cron-jobs to disable auto_rx, and enable those other decoders.

PR #330, opened 2020-12-18
Persistent APRS, DFM Frame ID Changes

darksidelemm:
WARNING: This makes some big changes to DFM frame IDs, which may cause interesting issues on the sondehub map when old & new stations upload positions for the same sonde.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ instructions](https://github.com/projecthorus/radiosonde_auto_rx/wiki).
 
 ### Contacts
 * [Mark Jessop](https://github.com/darksidelemm) - vk5qi@rfhead.net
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -68,7 +61,7 @@ instructions](https://github.com/projecthorus/radiosonde_auto_rx/wiki).
 
 ### Contacts
 * [Mark Jessop](https://github.com/darksidelemm) - vk5qi@rfhead.net
-* [Michael Wheeler](https://github.com/TheSkorm) - git@mwheeler.org
+* [Michaela Wheeler](https://github.com/TheSkorm) - radiosonde@michaela.lgbt
```
TheSkorm, line 64: 🥰

PR #358, opened 2021-01-12
Update index.html

Piter-NS:
Support for radiosondy.info when set to disable sending data to Habitat. 

When sending data to Habitat is disabled. In index.html file, I added support after clicking on sonde ID to redirect to radiosondy.info website for sonde details.

### DIFF
--- a/auto_rx/autorx/templates/index.html
+++ b/auto_rx/autorx/templates/index.html
@@ -249,7 +249,11 @@
                             if(autorx_config.habitat_enabled ==true){
                                 sonde_id_data.id = "<a href='http://sondehub.org/" + sonde_id + "' target='_blank'>" + sonde_id + "</a>";
                             } else {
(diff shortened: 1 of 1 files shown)


snh (changes requested):
Thanks for the contribution @Piter-NS ❤️ 

It'd be worth making the link conditional on `aprs_enabled` being `True` and `aprs_server` being`radiosondy.info`.

darksidelemm:
+1 on above. The extra conditions would be good in case users are pushing data directly into APRS-IS.

Piter-NS:
What is the APRS-IS server address?  After setting up the APRS-IS server, I could make the link go to aprs.fi maps.

darksidelemm:
Good idea to use an aprs.fi link if radiosondy.info is not used (and Habitat is disabled... which i kind of hope is never the case by the way...)
Unfortunately there's a lot of them, and most people would be using one in their country. I'd say if the aprs_server value is anything other than radiosondy.info, then use aprs.fi maps.

darksidelemm:
Thanks for making those changes, all good!

Issue #364, opened 2021-01-24
RS92 - GPS Ephemeris Downloading via FTP Broken


darksidelemm:
It looks like CDDIS closed off access via anonymous FTP in October 2020 (refer https://cddis.gsfc.nasa.gov/ ). Access is now only via HTTPS or ftp-ssl, both of which require a login (freely available, but would need to be added to the config somewhere).

This issue only affects decoding of RS92 sondes, which are obsolete and are hopefully on their way to being phased out.

Almanac downloading still works, though will result in a less accurate position solution.

mjgardes:
I got around that by changing to the esa server and update the file extensions.  I wanted to improve on that bit of code, but haven't got around to it yet.

The NWS offices near me are still launching RS92-NGPs and I don't see any indication that they or the other 18 or so NWS offices using the NGPs are looking into new models yet.  The US is pretty slow to adopt anything new.
```

  def get_ephemeris(destination="ephemeris.dat"):
    """ Download the latest GPS ephemeris file from the CDDIS's FTP server """
    try:
        logging.debug("GPS Grabber - Connecting to GSFC FTP Server...")
        ftp = ftplib.FTP("gssc.esa.int", timeout=10)
        ftp.login("anonymous", "anonymous")
        ftp.cwd("gnss/data/daily/%s/brdc/" % datetime.datetime.utcnow().strftime("%Y"))
        file_list = ftp.nlst()

        # We expect the latest files to be the last in the list.
        download_file = None
        file_suffix = datetime.datetime.utcnow().strftime("%yn.gz")

        if file_suffix in file_list[-1]:
            download_file = file_list[-1]
        elif file_suffix in file_list[-2]:
            download_file = file_list[-2]
        elif file_suffix in file_list[-3]:
            download_file = file_list[-3]
        elif file_suffix in file_list[-4]:
            download_file = file_list[-4]
        elif file_suffix in file_list[-5]:
            download_file = file_list[-5]
        else:
            logging.error("GPS Grabber - Could not find appropriate ephemeris file.")
            return None

        logging.debug(
            "GPS Grabber - Downloading ephemeris data file: %s" % download_file
        )

        # Download file.
        f_eph = open(destination + ".gz", "wb")
        ftp.retrbinary("RETR %s" % download_file, f_eph.write)
        f_eph.close()
        ftp.close()

        # Unzip file.
        os.system("gunzip -q -f ./%s" % (destination + ".gz"))

        logging.info(
[...]

PR #370, opened 2021-01-26
Testing

mjgardes:
Changing ftp server for [ephemeris file download for RS92](https://github.com/projecthorus/radiosonde_auto_rx/issues/364)

### DIFF
--- a/auto_rx/autorx/gps.py
+++ b/auto_rx/autorx/gps.py
@@ -14,7 +14,7 @@ import os
 def get_ephemeris(destination="ephemeris.dat"):
     """ Download the latest GPS ephemeris file from the ESA's FTP server """
     try:
-        logging.debug("GPS Grabber - Connecting to EDS FTP Server...")
+        logging.debug("GPS Grabber - Connecting to ESA's FTP Server...")
         ftp = ftplib.FTP("gssc.esa.int", timeout=10)
         ftp.login("anonymous", "anonymous")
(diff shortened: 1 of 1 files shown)


darksidelemm:
Thanks for this! I've run it up manually and it grabs the data. I don't have any RS92s to test with, so if you say it works with sondes in the air, I'll trust you on that!

darksidelemm:
@mjgardes It's looking like this change isn't always working. E.g. right now I just tried to run this and got the error:
```
DEBUG:root:GPS Grabber - Connecting to ESA's FTP Server...
DEBUG:root:GPS Grabber - Downloading ephemeris data file: brdc0750.21n.gz
ERROR:root:GPS Grabber - Could not download ephemeris file. - 550 File not found
```
Others are reporting similar issues. 

I suspect this may be because your solution is requesting a specific file - if ESA hasn't uploaded this file for whatever reason, then it breaks. My old solution would list the available files, and then pick the most recent one, so at least you would end up with a file that is still usable.

Is there a chance you can have a look into this?

mjgardes:
Interesting.  ESA's brdc directory is a few days behind and the ephemeris information seems to go stale pretty quick so that might be about as useful as letting it fall back to almanac-only mode.

However, the daily folders are still current.  I'll try grabbing the latest from there.

darksidelemm:
I've just pushed a change which uses the daily directory instead of the brdc directory. Hopefully this is reliable.

PR #410, opened 2021-03-23
Change HTML to show sondehub cards link once Radiosonde is "old"

LukePrior:
I couldn't figure out how to modify the backend of the program to show historical records so I made this instead which shows the historical link once the radiosonde is "old" (determined by rx_timeout). It is useful if you have the webpage to save information for several days. 

image

PR #411, opened 2021-03-23
Change HTML to show sondehub cards link

LukePrior:
Should fix #410


jasonmid:
IMO it would be nicer if the Sonde ID alone was a hyperlink to Sondehub rather than adding additional text to the end of it.

darksidelemm:
That probably is a better idea... Make it a link to the live sondehub page if the sonde is in the air, or link to the card page if its old. @LukePrior could you make this change?

PR #413, opened 2021-03-24
Update index.html

LukePrior:
This should be the changes you requested, please test.

### DIFF
--- a/auto_rx/autorx/templates/index.html
+++ b/auto_rx/autorx/templates/index.html
@@ -248,9 +248,9 @@
                         // Add a link to sondehub.org or radiosondy.info if enabled.
                         if (autorx_config.habitat_enabled == true) {
                             if (sonde_id_data.age != "old") {
-                                sonde_id_data.id = sonde_id + " (<a href='http://sondehub.org/" + sonde_id + "' target='_blank'>live</a> - <a href='http://sondehub.org/card/" + sonde_id + "' target='_blank'>historical</a>)";
+                                sonde_id_data.id = "<a href='http://sondehub.org/" + sonde_id + "' target='_blank'>" + sonde_id  + "</a>";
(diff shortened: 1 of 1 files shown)


LukePrior:
I'm working on redesigning the entire interface not sure if I will finish it but here is what I've got. Any suggestions welcome!

image

image

darksidelemm:
Hi Luke,

So I had been thinking about how i'd revise the auto_rx interface if I was going to do it. 
My intent was to bring it a bit more into line with the chasemapper interface. That is, a full-screen map, with data brought in with various tabs and controls that can be shown or hidden depending on what the user wants to see.
For example, the log could possibly be hidden by default, and opened up on demand. The SDR status (scanning, etc) would probably want to be displayed all the time somehow. The per-SDR information could end up being some kind of display that is added to the map (one per sonde) instead of a big table. This would make addition of data fields on a per-sonde basis a bit easier - e.g. dont even show the data if it isn't available (e.g. humidity, temperature, pressure). I would have pointed you at the wetterson.de interface as an example of that, but that seems to have been shut down.

As for making it work on both desktop and mobile devices... that's a big challenge to get right. Chasemapper barely works on devices with smaller than 7" screens, and I'm not sure I want to put in the extra effort to make it work well.

Cheers,
Mark

LukePrior:
I settled on a design for the mobile version fairly happy with it.
Screenshot_2021-03-25-12-00-26-89
Screenshot_2021-03-25-12-00-22-27
Screenshot_2021-03-25-12-00-17-30

LukePrior:
Is the time value passed through to the website in UTC?

darksidelemm:
I think only time-of-day (HH:MM:SS, as a string) is passed through to the web interface, and yes, it will all be in UTC.

PR #420, opened 2021-03-28
Redesigned Website

LukePrior:
I have developed a new interface for radiosonde_auto_rx from the ground up. My main goals for the new interface were customisation options, mobile support, and dark mode. I have also updated most of the libraries used to the latest versions for security and stability reasons. I have been running the build for a few days and it seems to be fairly stable but if you find any errors please let me know. It's my first time doing a project like this so my code isn't the neatest or fastest but I have tried to comment it where possible.

If anyone has time I would love it if you could test it out (make a backup first) and let me know what you think.


darksidelemm:
This looks really good Luke, awesome work! Much cleaner than my attempt, and the dark mode is nice!

A few questions / suggestions:
- When I load the page with no sondes being tracked, the 'time' column shows 'Invalid Date'. I'm guessing this can probably be inhibited somehow.
- Time/date field: Make this closer to ISO format time, e.g. 2021-03-26 00:00:01Z 
- For the settings pane, I'd suggest making the text a bit smaller, and having the button to the right of the setting name if possible. This may reduce the need to scroll so much in that pane (it also wasn't completely obvious that I had to scroll when I first looked at it).
- I see you're using colour coding to show the log level (black for <= INFO, red for everything above that). It's probably still worth including the level in the text, to differentiate between errors, warnings, and critical issues. 
- Does the upgraded socket.io mean we are compatible with python-socketio 5.x now? (I don't think it does, just wasn't sure)
- Not so sure I'm keen on the 'Voyager' map layer - it doesn't quite have enough contrast to my eyes. I think the original OSM map was better in that regard. 
- When going to dark mode (an in particular with the dark matter tiles), we probably need a different icon for the rx station, and maybe even for the balloons.

Anyway, really good job with this!

darksidelemm:
Another thing that might be worth looking into is the control endpoints I added a while ago. Note that I haven't got these enabled by default as I have concerns about security - Lots of auto_rx users have their web interface open to the internet and if we're going to enable control, then it means we'll need to add some kind of authentication... 

Anyway, the endpoints are here: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/web.py#L277
and you need to set this config value to enable them: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/station.cfg.example#L376

LukePrior:
Hi @darksidelemm 

In response to your questions/suggestions:

- If there are no tracked sondes the time column will now display nothing instead of 'Invalid Date'.
- I personally like the current time format but have added a toggle to display UTC IS08601 time in the table and logs if wanted.
- I have added the log level to the table but I'm not sure on the formatting of it, suggestions welcome.
- The socket.io version upgrade was just to 1.7.4, I tried 5.x but that would require significant changes so gave up.
- The default map layer has been changed to OpenStreetMaps Mapnik, I'm using the Leaflet-providers plugin to get the layers. You can view available layers at the preview [here](https://leaflet-extras.github.io/leaflet-providers/preview/index.html), not all are compatible only those that don't require authentication. Please provide any feedback about what maps to add, remove, etc.

I will work on redesigning the settings menu and getting new icons for when in dark mode. I have pushed my changes above for you to try in the meantime.

LukePrior:
I have added dark mode icon support currently, it only adjusts the home icon but I think this should be fine.

darksidelemm:
Dark mode icon looks good!

I'd suggest if you're going to use local time in the time field, then maybe change the title on the table to indicate this (e.g. Time (Local) vs Time (UTC), otherwise it will be a bit confusing. When dealing with radiosondes I think most people expect UTC time data, as radiosonde data is used internationally, so maybe UTC as a default would be good.

I'm not sure if this is possible, but on the table can some of the fields be narrowed up a little? e.g. RH, Az, El are probably a little bit wider than they need to be, resulting in some fields falling off the right of the screen.

LukePrior:
I have changed UTC time to be the default and made local time an opt-in option, I can't modify column names on the fly so this solution will have to suffice. I have modified the table code so hopefully it should size better.

LukePrior:
I'm having a bit of trouble changing the settings menu styling, so it seems like the best I could do is show the scrollbar?

darksidelemm:
Can the font size not be decreased?

LukePrior:
I just reduced it from 25px to 20px have a look

LukePrior:
That should help improve things in the setting tab

darksidelemm:
Hmm, not seeing a huge difference here? 
The formatting i was looking for was a bit more like how the settings tab on the habhub tracker works
(e.g. visit https://tracker.habhub.org/#!mt=roadmap&mz=5&qm=1_day&mc=53.46751,-2.23389&q=!RS_*; and click the gear icons on the top right)

LukePrior:
Try that, I'm not sure it looks better but it is more space-efficient. The CSS seems to cache for quite a long time so you might not notice smaller text.

darksidelemm:
Yep, I can see that change. I guess the only improvement now would be to left-align the text, and right-align the button. Other than that, looking good!
I also see the table column size changes, that works fine!

darksidelemm:
OK, the smaller font is now showing up for me!

LukePrior:
I aligned the text with the table above and made the buttons a bit smaller so it should all fit ok.

darksidelemm:
Yep this looks good! I'll merge this into the testing branch now. 
I have a few other changes relating to sondehub that I want to get out the door soon, so this will get released as part of 1.5.1.

darksidelemm:
Thanks again for this work Luke! 

There's certainly more that can be done with the web interface - being able to add control over the scan process, and being able to manually start up a decoder would be a good addition, though this does mean adding some form of authentication.

F1TKE:
Hi, i test this new website, with scan plot actived, no frequency information for the plot with the mouse of the plot. Nice work.

LukePrior:
> Hi, i test this new website, with scan plot actived, no frequency information for the plot with the mouse of the plot. Nice work.

Hey,

I'm glad you like the new website, please give any suggestions/feedback. I had to disable the mouse over information for the scan chart as the formatting was broken and it didn't work on mobile. Feel free to re-enable it in scan-chart.js and try for yourself.

darksidelemm:
I'd suggest enabling it again. I'm not really sure mobile users are going to be looking at the scan plot that much anyway?

LukePrior:
> I'd suggest enabling it again. I'm not really sure mobile users are going to be looking at the scan plot that much anyway?

I will do some testing and see what I can do.

F1TKE:
i have enable the show, effective don't work fine. disable it again

image

image

LukePrior:
Yeah that was the same issue I was having when trying to get it working, will have another look at the CSS when I have time. If you can find what CSS values (Chrome DevTools) need changing to make the tooltip display correctly let me know and I can get them changed.

mikethebee:
I just had updated 'testing' and believe I see the new website. It is looking good, and seems faster. Running in FF 87 on Ubuntu. An initial thought is that the map keeps self centering, which I don't recall happening previously. I would like to have an option to disable this, so I can centre the map manually for tracking a distant balloon with a larger zoom setting.

LukePrior:
> I just had updated 'testing' and believe I see the new website. It is looking good, and seems faster. Running in FF 87 on Ubuntu. An initial thought is that the map keeps self centering, which I don't recall happening previously. I would like to have an option to disable this, so I can centre the map manually for tracking a distant balloon with a larger zoom setting.

Have you tried disabling Follow Sonde in the settings menu, this should stop the map from moving.

mikethebee:
> > I just had updated 'testing' and believe I see the new website. It is looking good, and seems faster. Running in FF 87 on Ubuntu. An initial thought is that the map keeps self centering, which I don't recall happening previously. I would like to have an option to disable this, so I can centre the map manually for tracking a distant balloon with a larger zoom setting.
> 
> Have you tried disabling Follow Sonde in the settings menu, this should stop the map from moving.

Ah, thanks,  I had missed the 'below the fold' options, need to scroll down by mouse as I don't see an elevator. (I am on a desktop.)

LukePrior:
Yeah that seems to be a common sticking point, I will have a think about how to make it a bit more Intuitive that you can scroll down.

mikethebee:
Maybe a 'down arrow' at the bottom.

mikethebee:
The 'table option' list could be collapsed. I don't think I would be referring to it as regularly as I would the other options.

LukePrior:
@darksidelemm 

When I was updating the website I was having a look at how ChaseMapper was configured and I noticed that you implemented landing predictions with the Tawhiri API. I know you said you intend ChaseMapper for tracking but if it's similar enough I think it would be appreciated by people who like having everything in one place dashboard style. The code appears to be similar but I was wondering if it's compatible or easy to port over?

LukePrior:
@mikethebee 

Thanks for the suggestions, I'll look into them.

darksidelemm:
I'd prefer to leave predictions out of the main auto_rx interface. I would prefer to just link users to the sondehub tracker, where they can get predictions, *and* see data from other receivers.

whallmann:
New Design looks good, but there are some more details to work on.
Mostly i access on my mobile phone. Seems the visible screen is all we have - no scrolling possible, thats bad - to less place on this. Why is vertical scrolling blocked?
Also the selected colums for the list on top are los after next page view. The minimum starter columns are not that what i like to see. What about a safe button to place a cookie for that config.  On the settings page the selection of availble columns are to small for mobile use, cant touch them with my fingers. 
Will not do screem as much, the idea of the new the design is worth to go one further. thanks for that guys.
cu
Wolf

PR #425, opened 2021-04-02
Website Interface Improvements

LukePrior:
I have made a number of improvements to the web interface:

- table updates every second so age field looks nicer
- removed watercolour theme
- enabled dark mode station icon when WorldImagery map is selected
- UTC Chart now follows UTC time display preference
- Hyperlinks in data table are no same colour as other but bold
- I have added the tooltip back to the graph
- The scan chart now displays better on mobile without overlapping x-axis tags
- If no sonde data is available a placeholder message is shown
- Added a feature when row in table is right clicked or long pressed on mobile the corresponding sonde on the map will be highlighted
- Created new balloon icons for a total of 6 colours up from 4
- Tabulator no longer resets view on data update for better experience if column size changed or scrolled on mobile
- The number of rows to display per page is now an exposed option in the interface (default 6 desktop, 3 mobile)
- Slightly improved the settings menu display by enabling flex display

These improvements should bring the interface to a point where it is ready for release, please test and let me know of any issues.

I am having trouble finding a good solution (code wise) for indicating scroll in the settings tab, if anyone wants to give it a shot please do.


darksidelemm:
Looking good! I'm planning on rolling out the socketio updates tomorrow when I'm a bit more awake - this will be 1.5.1-beta2. I'll also have to update chasemapper at the same time, since in most cases people are running both on the same machine (and hence same dependencies). Nice thing is the socket.io.min.js file is just a drop-in change.

darksidelemm:
Hmm, we seem to have lost a lot of the nice detail in the balloon and parachute icons with this update. They looked pretty nice before, now not so much? 

balloon-purple

balloon-red

LukePrior:
> Hmm, we seem to have lost a lot of the nice detail in the balloon and parachute icons with this update. They looked pretty nice before, now not so much?
> 
> Old example: balloon-purple parachute-purple
> 
> New example: balloon-red parachute-red

Yeah it seems my editor didn't really work that great, maybe best to revert to the originals?

If anyone is good with graphics and wants to make some icons I would be more then happy to get them working but I'm a bit hopeless in the art department.

PR #429, opened 2021-04-03
Small Website Bug Fixes

LukePrior:
I've got a few general website bug fixes here:

- Change Station name display position so that it looks correct on mobile and desktop
- Fixed a small bug that prevented table column display preferences from getting saved
- Disable the placeholder No Sonde Data message as it would break in certain situations
- Disable the changing of icon colour to white for balloon as icons were changed back

I was also wondering if you knew anything about a `Error on request: %s` message in the logs and what it might mean.

Cheers


darksidelemm:
Hmm unsure about the 'error on request' messages, I'm wondering if they are coming from EngineIO or similar.

LukePrior:
> Hmm unsure about the 'error on request' messages, I'm wondering if they are coming from EngineIO or similar.

It appears to just be a socket.io problem, any idea if it's something that needs fixing?

image

darksidelemm:
Oh, that's thrown whenever you close your browser tab. 
Its possible there need to be some kind of clean shutdown/disconnect done on the browser side, though I'm not sure how that works...

The other option is to inhibit those errors somehow, but that may hide other errors which we do want to see...

LukePrior:
> Oh, that's thrown whenever you close your browser tab.
> Its possible there need to be some kind of clean shutdown/disconnect done on the browser side, though I'm not sure how that works...
> 
> The other option is to inhibit those errors somehow, but that may hide other errors which we do want to see...

A good compromise might be to make the error message displayed a bit more descriptive if possible?

darksidelemm:
Perhaps, but this error is thrown at a much lower level than we really have access to. It would mean doing something with the EngineIO error handler, possibly even a change to that codebase, which is not really something I want to be doing...

darksidelemm:
From this issue ( https://github.com/miguelgrinberg/python-engineio/issues/83 ) there's this comment:
```
The session is disconnected error is not something you can catch yourself. The problem occurs when a client does not timely send ping packets to the server, as indicated by the Socket.IO protocol. A client should send a ping packet every 25 seconds (in the default configuration). If the server does not receive a ping in 60 seconds it declares the client gone and closes that connection. If after 60 seconds, the client comes back and tries to send a ping you'll get the "session is disconnected" error, because the server already gave up on that client and fully released it.
```

I suspect what may be going on here is when a browser tab goes into the background on modern browsers (Chrome, etc), it gets massively rate limited, to the point where a lot of JS stops running. So those pings are never sent. When you go back to the tab things start to run again, and by that point the server has dropped the session, so the client has to reconnect. Not sure what the solution to that is.

LukePrior:
Yeah that must be itonly happen on certain devices/browsers, looks like there isn't much we can do.

F1TKE:
hello
with 1.5.1 beta2 last version, i can only use Chrome. 
With another navigator (Chromium, firefox, IE)=>no log, no statut for current task, no waterfall

auto-rx log say:
Apr 03 17:33:06 F1TKE-RS auto_rx[12082]: The client is using an unsupported version of the Socket.IO or Engine.IO protocols (further occurrences of this error will be logged with level INFO)
Apr 03 17:33:06 F1TKE-RS auto_rx[12082]: 2021-04-03 17:33:06,273 ERROR:The client is using an unsupported version of the Socket.IO or Engine.IO protocols (further occurrences of this error will be logged with level INFO)

same with windows , linux, android 

turn RPI3 with BUSTER

darksidelemm:
First, make sure you have updated flask-socketio to the latest version. (`sudo pip3 install -U flask-socketio`)

Then, you will need to clear your browser cache on whatever browser you are using to ensure you have the latest socketio JS version. Otherwise you will get the error you have mentioned.

F1TKE:
all work again, continuous testing

PR #431, opened 2021-04-04
Using venv and Code quality improvement

F4HVQ:
Some commits to:
- use virtual python environment (venv). Wiki should be update 
- improve code quality in utils folder


darksidelemm:
Woah woah woah, threse are a *lot* of changes, and you're doing it to the master branch. The testing branch is already ahead a bit by now, with changes to the python versions supported, amongst other things.

I'm happy to accept some of the changes to the utilities (re-targeted to the testing branch), but a lot of your other changes needs to be taken slowly, one at a time. The venv change in particular needs a lot of testing and changes to the documentation. I'm actually moving towards suggesting the use of Docker for users new to Linux, as it's considerably simpler to setup, and handles all the dependency issues (and updating) easily.

Other comments:

- .gitignore file changes - Good! And probably should have been done a while ago.
- Pinning of python requirements should no longer be required as of 1.5.1. I have updated socket.io and now the latest flask-socketio will work (and you *need* to use the latest now, v5.0.1 at a minimum). 
- A lot of your new utils are using a shebang line of `#!/usr/bin/python3.5` - does this mean you've been developing on a machine with only Python 3.5? In the testing branch I've now dropped support for Python 3.5 - Python 3.6 is now the minimum version supported. This was required to be able to update the flask-socketio library, and so I can start using f-strings throughout the project (much neater than using .format() )
- I see that you are using some code in the libs directory in a few of the utils files. Are these libraries *Really* necessary? I see you are using your own CSV handler and directory handling routines. I would prefer to be using python built-ins, even if it means a few lines are a bit more convoluted (e.g. using glob). You've added a lot of seemingly un-necessary library into the auto_rx project, along with a lot of unit tests which I don't think should be in this repository.

I appreciate this was probably a lot of work, but it would have been good if you'd made some contact first to discuss the changes!

F4HVQ:
Hi @darksidelemm,

"you're doing it to the master branch" : On Github interface I did see testing branch

"your new utils are using a shebang line of #!/usr/bin/python3.5" : You are right this libs files come from a other project. But I think you can use it with newer Python version.

"I see that you are using some code in the libs directory in a few of the utils files." : I prefer create simple functions that I can reuse. It's not a good practice to rewrite code. We can introduce errors. For example : getting the list of files or folder is always the same code.

"but it would have been good if you'd made some contact first to discuss the changes": Yes, but I didn't find the GitHub menu to email you.


Very sorry for all theses problems. I hope you can refuse this pull request.
If it is not possible can I help you to fix this ?

darksidelemm:
> "you're doing it to the master branch" : On Github interface I did see testing branch
The testing branch is what all PRs should be against. I realise this wasn't clear, but if you'd discussed the changes, or submitted a smaller PR then this would have been made clear.

> "your new utils are using a shebang line of #!/usr/bin/python3.5" : You are right this libs files come from a other project. But I think you can use it with newer Python version.
Sure, it would be fine with anything newer, but specifying a particular python version in the shebang probably isn't best practice.

> 
> "I see that you are using some code in the libs directory in a few of the utils files." : I prefer create simple functions that I can reuse. It's not a good practice to rewrite code. We can introduce errors. For example : getting the list of files or folder is always the same code.

Sure, but when you're just replicating functionality that already exists in the python standard libraries (e.g. glob, or csv reading), then I don't see the point. I don't see a problem re-using few short lines of code, given that you're replacing it with a rather large amount of code that is essentially a replication of stdlib functions. 

> "but it would have been good if you'd made some contact first to discuss the changes": Yes, but I didn't find the GitHub menu to email you.

My e-mail has been on the main page for auto_rx for a very long time now, and you can always comment on an issue to get our attention.

I'm happy to accept contributions, but they should be more targeted than this. Perhaps a PR to update some of the utils script would be a good start. I'm sure there's improvements to be made there.

PR #447, opened 2021-04-18
Iss443 - update global dictionary based on GPSD if there has been a significant change in location from station.cfg

g7gpr:
Had some strange effects with concurrency when a GPSD update occurred during a position upload, which is why the extra checks have been added to the habitat uploader.

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -612,6 +612,24 @@ def station_position_update(position):
     # Quick sanity check of the incoming data
     if "valid" not in position:
         return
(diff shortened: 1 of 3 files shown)


darksidelemm:
Could you please re-target this into the testing branch. There have been numerous updates, and there are also new position upload functions in the sondehub uploader.

To be honest the GPSD stuff is way more trouble than it's worth. The *only* use-case where I see it being useful is if someone is running an az/el tracking system 'mobile', and I don't know if anyone is actually doing this. A fixed station has zero need to use GPSD, and a 'normal' mobile station should be uploading their chase position information using chasemapper, not auto_rx.

PR #449, opened 2021-04-20
Change Update Link if beta release

LukePrior:
This is a bit more of a personal preference change but I don't believe it negatively impacts anything.
I like to look at the changes with each testing beta version and as no changelog is provided I believe linking to the commit history is best.


darksidelemm:
I'm OK with this.

PR #451, opened 2021-04-21
Interface Design For Control Modal + Unread ERROR log notification

LukePrior:
**This should not be merged yet.**

This is the style for the web interface, it works decently well with dark/light mode and desktop/mobile screens.

The buttons do nothing they need to be hooked up to the backend functions.

image


LukePrior:
I have added my changes to highlight the tray icon red when there are unread error logs

PR #455, opened 2021-04-23
Fixes for web and mobile

LukePrior:
- Text in control modal now matches theme always
- viewheight now works better on mobile and desktop
- switch from cookies to localstorage
- password is stored in localstorage
- auto login if saved
- sidebar only red for error and critical logs


LukePrior:
There are still fixes and improvements which can be made but I think this is in a good enough state for 1.5.2.

PR #457, opened 2021-04-24
Control Buttons disable until ready to issue new command

LukePrior:
I've implemented the changes discussed, buttons are disabled once the command has been executed until the current task is updated or 10s has elapsed whichever is first.t


LukePrior:
Found a bug, fixing

LukePrior:
Ready to merge

PR #458, opened 2021-04-24
Additional data displayed in the tooltip

Piter-NS:
Removed the double tooltip on the map and added additional data displayed in the tooltip, such as altitude, climb and horizontal speed.
A style has been added for the sondeTooltip class.

### DIFF
--- a/auto_rx/autorx/static/css/main.css
+++ b/auto_rx/autorx/static/css/main.css
@@ -448,3 +448,8 @@ select[disabled]{
         background-color: #414141;
     }
 }
+
+.sondeTooltip {
+    font-size: 11px;
+    font-weight: bold;
+}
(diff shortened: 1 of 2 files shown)


darksidelemm:
Thanks!

darksidelemm:
I just noticed the tooltip is always visible. This is probably fine, but I'll have a play with it during an actual flight and evaluate :-)

LukePrior:
Is the tooltip also always visible on mobile because I can imagine that this it would get quite annoying, I think the previous functionality was ideal.

LukePrior:
This also won't work for those who have imperial units enabled.

darksidelemm:
Yeah all good points. It looks like it will behave the same on mobile as on desktop.
It actually doesn't look to bad on mobile. It's small enough that it doesn't cover too much. 

Screenshot_20210424-205620


Definitely needs to deal with both metric and imperial units though, and the altitude needs to be converted to fixed point in metres (currently its showing two un-necessary decimal points).

LukePrior:
> It actually doesn't look to bad on mobile. It's small enough that it doesn't cover too much.

Will it disappear once the sonde is marked as old because if you have multiple it will get messy.

darksidelemm:
Yeah, it won't. So yep, I think this needs to go back to being a hover-over thing.

PR #459, opened 2021-04-24
Add imperial support to tooltip, make tooltip non-persistent.

darksidelemm:


LukePrior:
Should the rounding of values for the tooltip be equal to those used in the table?

We have 1 decimal place for altitude and velocity in the table but 0 in the tooltip so the data will not match.

It's quite nitpicky of me so if you think the saved space in the tooltip is worth it keep it as is.

darksidelemm:
I figure 1m resolution for altitude is sufficient, same for ground-speed (horizontal velocity). Ascent rate deserves a little extra precision.

PR #467, opened 2021-04-30
Hack around subpixel issue to make it less noticeable

TheSkorm:
This is the workaround discussed in this leaflet issue to make the grid lines disappear on high DPI monitors. While it doesn't resolve the issue fully, it is MUCH better

https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-688644225
**Before**
Screen Shot 2021-04-30 at 11 13 23 362
**After**
Screen Shot 2021-04-30 at 11 13 16 702

### DIFF
--- a/auto_rx/autorx/static/css/main.css
+++ b/auto_rx/autorx/static/css/main.css
@@ -453,3 +453,9 @@ select[disabled]{
     font-size: 11px;
     font-weight: bold;
 }
+
+/* workaround for subpixel lines https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-688644225 */
+.leaflet-tile-container img {
+    width: 256.5px !important;
+    height: 256.5px !important;
+}


LukePrior:
I'm seeing this issue on my 27" 1440P monitor the lines seem to flash each time the map moves to follow the sonde.

LukePrior:
https://github.com/Leaflet/Leaflet.TileLayer.NoGap

argilo:
This workaround causes the map to be noticeably blurry (on my low-DPI monitor, at least).

darksidelemm:
I think we already have NoGap included now?

darksidelemm:
Nevermind, getting confused with sondehub tracker. So no, we aren't using NoGap here. Maybe we should be?

argilo:
Aha, that might explain why I'm not seeing blurry maps on sondehub-tracker.

argilo:
Leaflet 1.9.4 claims to have a fix for this: https://github.com/Leaflet/Leaflet/releases/tag/v1.9.4

argilo:
That fix doesn't seem to work, but NoGap does. I opened a pull request: #925

PR #474, opened 2021-05-01
Visual Studio Code Dev Container

snh:
Can also be used for local development without Visual Studio Code.

Updates to https://github.com/projecthorus/radiosonde_auto_rx/wiki/Development-&-Contributing incoming once this merges.

/cc @LukePrior @darksidelemm


LukePrior:
Did you get a chance to update the wiki page?

I'm still running the normal beta docker version but would be interested in switching.

PR #476, opened 2021-05-01
Smooth Scroll Plugin

LukePrior:
I'm not sure if this provides a better experience. 

Give it a go and let me know.


LukePrior:
I've been trying this for a bit bit I don't think it's that much better.

PR #493, opened 2021-05-15
Update buttons, home button, plot landings

LukePrior:


LukePrior:
CSS fix needed

LukePrior:
Good now

PR #500, opened 2021-05-18
Add option to export log files from web interface.

Moll1989:
Adds a button in the settings menu to export all files in the auto_rx log directory.
This allows a user to conveniently access telemetry and system logs with limited Linux / SSH knowledge.
Logs are added to a zip file and a time stamp is added to the filename when the request is processed.

### DIFF
--- a/auto_rx/autorx/templates/index.html
+++ b/auto_rx/autorx/templates/index.html
@@ -1568,6 +1568,13 @@
                   </div>                            
                   <br>
                   <br>
+                  <h2 style="display:inline;vertical-align:middle;">Export Logs</h2>
+                  &nbsp;
+                  <div style="display:inline;vertical-align:middle;">
+                    <button onclick="window.location.href='/export_logs'">EXPORT</button>
+                  </div>                 
+                  <br>
+                  <br>
                   <br>
             </div>       
         </div>    
--- a/auto_rx/autorx/web.py
+++ b/auto_rx/autorx/web.py
@@ -7,12 +7,15 @@
 #
 import copy
 import datetime
+import io
 import json
 import logging
+import pathlib
 import random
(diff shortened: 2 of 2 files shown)


darksidelemm:
I would have preferred if you had discussed this PR somewhere first!

I'm not sure that zipping up all the log files and sending them all out at once is the best idea here. There are stations where that will result in auto_rx having to zip up hundred of mb of log files, and then serve them up. I suspect this might lock up the web interface for a while...

Also, your approach of zipping up all the files in the log directory also means that stations with the --system log option enabled will have those logs included, which adds to the download.

IMO a better option would be to add a way to download log files on the historical page, e.g. based on the selected sondes. This way users can select a subset of sonde logs to download, and then this is passed as an argument (by serial number) to the back-end, which zips up the appropriate log files and serves them.

LukePrior:
Yeah if the option could be added to the historical page so only selected sonde logs were downloaded that would be great.

darksidelemm:
As an addition to this, the method I'm using to find a particular log by serial here: 
https://github.com/projecthorus/radiosonde_auto_rx/blob/testing/auto_rx/autorx/log_files.py#L441

There's an example of passing arguments to the backend in a POST request here: https://github.com/projecthorus/radiosonde_auto_rx/blob/testing/auto_rx/autorx/web.py#L302
I'd suggest passing the list of serial number to the backend as a JSON-encoded list, however as these parameters are included in the URL, I'm unsure what the limit on this length is.

darksidelemm:
Turns out passing the argument list (list of serials) to the backend was a bit more complicated than I expected. Ended up having a go at this myself this morning, which has been merged in under https://github.com/projecthorus/radiosonde_auto_rx/pull/502

PR #527, opened 2021-06-21
Tweaks to historical table

darksidelemm:
Testing out the addition of minimum height and maximum range data. Sidebar has been expanded slightly, and I've switched to a shortened form of the sonde type, with a tooltip showing the full type on hover-over.

### DIFF
--- a/auto_rx/autorx/__init__.py
+++ b/auto_rx/autorx/__init__.py
@@ -17,7 +17,7 @@ except ImportError:
 # MINOR - New sonde type support, other fairly big changes that may result in telemetry or config file incompatability issus.
 # PATCH - Small changes, or minor feature additions.
 
-__version__ = "1.5.4-beta3"
+__version__ = "1.5.4-beta4"
 
 
 # Global Variables
--- a/auto_rx/autorx/log_files.py
+++ b/auto_rx/autorx/log_files.py
@@ -20,6 +20,7 @@ import numpy as np
 from dateutil.parser import parse
 from autorx.utils import (
     short_type_lookup,
+    short_short_type_lookup,
     readable_timedelta,
     strip_sonde_serial,
     position_info,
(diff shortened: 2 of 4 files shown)


fkemps:
Hi Darksidelemm,

Thanks for adding the min and max height columns on the historical overview.
For the min height we currently see the launch height. The request was actually to list the last and lowest height rx-received during landing. 

Can change to list the last lowest height rx-received height?

darksidelemm:
The current behaviour is:
```
                    if _output["last"]["range_km"] > _output["first"]["range_km"]:
                        _output["min_height"] = int(_output["last"]["alt"])
                    else:
                        _output["min_height"] = int(_output["first"]["alt"])
```

So... if the last received position is further away from the first received position, then its minimum altitude will be used. Either way, you get an indication of the minimum altitude your station received a signal at the furthest distance, which I think is a reasonable metric? The displayed maximum range value will correspond to the minimum altitude.

fkemps:
Hello,
If I read the code correctly, will the min_height be set to the highest value of "last" or "first".

When "last" > "first" then first should be used, else "last". Agree?

I do get your thoughts to use the lowest height, from launch or landing. #smart!

darksidelemm:
min_height is set to the altitude of the whichever point (first / last) is furthest away from the receiver station.

If it was as you suggested, then it would be using the closest point, and would probably always show the launch site, if the receiving station is close to it.

fkemps:
The case I try to address is shown in the image. Of interest is the lowest height during landing. That's used for retrieving the sonde. 

To address receiving close by launch site we could multiply _output["first"]["range_km"] * 4 for example.

auto_rx-history-m-height

darksidelemm:
The point of the metrics is to provide some indication of receiver performance. Showing the furthest and lowest signal does that, and I think it's good enough. If you want more information you can plot the path just as you are doing. 
I can't easily support all possible situations of receiver to launch and landing site geometry, as it will make the table huge...

PR #542, opened 2021-07-08
Added dashboard integration with SQ6KXY Radiosonde Tracker

fkemps:
Website SQ6KXY is used a lot with sonde hunters to observe their rx-uploads and sonde hunt status (Unknown,Found, Lost, Needs attention etc)

This change makes it very easy to pull that information on a newly opened page.
Reachable through clicking on the Frame counter value, as explained on the table hover-over.


darksidelemm:
I'm not sure putting the link to radiosondy on the frame field is the right place to put this... There needs to be a neater way.

darksidelemm:
Noting that there isnt much width available in the table, I think a better option would be to have a column which has small icons (one for sondehub, one for radiosondy), which link to the appropriate destination. Each icon should only appear if that particular uploader (sondehub or APRS) is enabled.

fkemps:
I tried first on the historical page but that was more complex.
Next to that, I would take way to many clicking around. Now strait from the main dashboard.

On radiosondy you can find your contribute received frames from auto_rx, used a lot for antenna receive (antenna) tuning.

fkemps:
Ooh sure dynamic icons will be better, agree.

darksidelemm:
TBH the entire table idea needs to be re-worked and turned into something a bit more similar to how Sondehub looks, but thats a lot more work. 

For Icons, I'm not entirely sure what would be best. The favicon.ico files from each relevant site might be suitable, but we'd have to see how they look in dark mode as well.

fkemps:
Think you underestimate the table value. Amongst others the table + plot is my day-to-day primary dashboard monitoring point.

Another option is to make ID right-clickable with a dropdown menu to select external resources.
That wouldn't cost any display space, not suffer from graphics dark/light-mode, and allows to scale the number of integrations.
I'm not a programmer but can give it a try "I never tried that before, so I should definitely be able to do it" (Pipi Longstocking)
(https://stackoverflow.com/questions/4909167/how-to-add-a-custom-right-click-menu-to-a-webpage)

This initial integration on testing branch was already much appreciated by a Dutch group of sonde hunters.

TheSkorm:
Don't add custom right click, it's horrible undiscoverable user experience

darksidelemm:
I would much prefer either a few small icons in the table somewhere (perhaps extend the ID field slightly to fit them), or a complete re-work of the page to the move the per-sonde information into a format a bit closer to how the sondehub tracker displays things. 

It's also worth noting that your link to Radiosondy is only going to work for RS41 sondes. 
Radiosondy is taking data via APRS, and so we have to munge the serial number into something thats fits within the APRS 9-character object ID lengths. RS41 serials thankfully fit into 9 character just fine, but everything else has to be re-formatted in various ways, which different from sonde-to-sonde. The conversions are currently done here: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/aprs.py#L39

This APRS-formatted callsign is currently never sent to the web interface, and is currently only created in the APRS Exporter class. To be able to make a radiosondy link work properly that conversion will have to be moved elsewhere, with the APRS callsign passed to the web interface somehow. This is doable, but it's a subtlety that your PR hasn't covered.

darksidelemm:
The changes you have made are still not right... 
This is going to require more changes than just in the front-end to make this work - I'm going to have to rearrange how the APRS callsigns are generated a bit. I'm going to take a look at this on the weekend, didn't get to it last weekend.

darksidelemm:
Also I would still be including iMet callsigns, as if auto_rx users are uploading data to radiosondy then there will be data they can look at. Not really my issue if other decoders decided to go generate a different callsign, especially since I had the algorithm to generate the callsign available publicly for a long time...

LukePrior:
I think the best way to do this might just to have a website settings toggle option to switch to radiosondy.info for links. You should be able to easily copy the existing code to create a new option.

darksidelemm:
Worth noting that in the latest testing branch there is an 'aprsid' parameter passed to the frontend in the telemetry, so it should be doable to add in the radiosondy link now. The parameter will be null if there is no callsign available (e.g. M20 right now)

darksidelemm:
We've already discussed the columns before... We've got enough columns as it is, adding more makes that table wide and wider and I don't really want that.

PR #548, opened 2021-07-21
Docker: Use Debian's Slim Buster as the base for the container image

snh:
Initial testing on a Pi Zero W seems to indicate that this solves the NumPy issue.


yaakov-h:
Confirmed as working on Zero W.

PR #554, opened 2021-07-24
1.5.5-beta4 - LMS6-1680 Updates, Web interface tweaks, M20 APRS uploading

darksidelemm:


fkemps:
I confirm both Icons next to serial and Last R column on history are working well.

PR #558, opened 2021-08-15
v1.5.5 release

darksidelemm:
* APRS: M20 telemetry can now be uploaded to APRS-IS (e.g. radiosondy.info)
* APRS: The recommended comment string has been changed to include the full sonde serial number. Refer to the configuration file change-log: https://github.com/projecthorus/radiosonde_auto_rx/wiki/Configuration-File-Changelog#version-155
* APRS: Removed the option to upload telemetry as position reports instead of APRS objects. The convention for telemetry uploading is to use Objects, and the position report option was causing more trouble than it was worth.
* Decoders: The LMS6-1680 decode chain has been improved, with a good few dB of improved sensitivity. There is also the option to enable an alternate decode chain which gives slightly higher performance again, which can be selected between in a new configuration file option (refer change-log linked above). However, this decode chain uses a considerable amount of CPU, so is only recommended for use on a RPI 3B+ or better.
* Scanning: Detection of LMS6-1680 sondes has been improved somewhat through the use of a wider FM filter bandwidth.
* Web Interface: Instead of the serial number in the table being a link, there are now small icons next to each serial number, linking to Sondehub and radiosondy.info.
* Historical Display: Table now shows Last Height, Maximum Height, and Maximum range for each sonde.
* Historical Display: The coverage map now sets any bearings with no sonde data to the station location and 0 range. This has the effect of making the coverage map a little 'spiky', but is a better representation of the data.
* Docker: Fixes to running auto_rx on the Pi Zero W (Note that the Zero W is right on the bare edge of having enough CPU grunt to run auto_rx).


F1TKE:
image

darksidelemm:
You're getting ahead of yourself a bit... this hasn't been merged into the main branch yet.

PR #563, opened 2021-08-17
Flush stdout only

snh:
Doing a `fflush` on `stdin` was preventing `fsk_demod` from functioning on Alpine Linux, which uses musl libc (https://wiki.musl-libc.org/functional-differences-from-glibc.html).

My understanding is that `fflush` is generally used for clearing the output stream, and when used on input streams is only really usable for seekable files, so in most cases is ineffective when used with `stdin`.  I'm not sure what the purpose of flushing `stdin` was here?

Also noticed it was alternatively checking for `fout == stdin`, which in theory should never be the case.

In my limited testing this fixed the issues I experienced on Alpine Linux with no obvious negative impact to glibc environments, though I lack context as to why it was the way it was, so may be missing something.

/cc @drowe67

### DIFF
--- a/utils/fsk_demod.c
+++ b/utils/fsk_demod.c
@@ -6,7 +6,7 @@
 
   C test driver for fsk_demod in fsk.c. Reads in a stream of 32 bit cpu endian
   floats and writes out the detected bits
-   
+
 
 \*---------------------------------------------------------------------------*/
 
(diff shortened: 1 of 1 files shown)


darksidelemm:
I suspect this will probably be fine, but it will definitely need testing on multiple platforms. That's what the testing branch is for! 
Apologies to those users affected if this breaks something! (Unlikely though)

snh:
This mornings Sonde release was picked up by my Debian based container running this patch OK, so far so far good, but will be good to get more exposure once in `testing`.

I'll be running my Alpine based container for the next few days to make sure it is OK too.

https://stackoverflow.com/questions/18170410/what-is-the-use-of-fflushstdin-in-c-programming has some interesting discussion on the merits of `fflush` with `stdin`.

PR #565, opened 2021-08-20
Docker: Switch to Alpine Linux

snh:
Results in a container that should be a little leaner, and a little faster to build.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -18,11 +18,14 @@ jobs:
 
       - name: Calculate Container Metadata
         id: meta
-        uses: crazy-max/ghaction-docker-meta@v1
+        uses: docker/metadata-action@v3
         with:
           images: ghcr.io/${{ github.repository }}
-          tag-semver: |
-            {{version}}
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=ref,event=branch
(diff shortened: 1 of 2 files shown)


snh:
Before merging, I need to confirm that the following warning when compiling on 32bit architectures isn't going to pose a problem.

```
fsk_demod.c:359:45: warning: format '%ld' expects argument of type 'long int', but argument 3 has type 'time_t' {aka 'long long int'} [-Wformat=]
  359 |                 fprintf(stderr,"\"secs\": %ld, \"EbNodB\": %5.1f, \"ppm\": %4d,",seconds, stats.snr_est, (int)fsk->ppm);
      |                                           ~~^                                    ~~~~~~~
      |                                             |                                    |
      |                                             long int                             time_t {aka long long int}
      |                                           %lld
```

snh:
Running in to https://github.com/moby/moby/issues/40734 on 32bit architectures with this one too, so we'll need to put the brakes on this PR for now until we wait for Raspbian to update `libseccomp2` or an alternative easy to implement solution is available.

Technically the container in this PR is fine, and the problem is with the host OS, but we don't want to introduce too much friction if we can help it.

https://docs.linuxserver.io/faq#libseccomp is a good write up of some alternative solutions, but I'm a little nervous about how many people would need to implement these if we move to Debian Bullseye of Alpine 3.13+.

Issue #585, opened 2021-11-13
Failure to start systemd service when there's no network to allow version checking


dfgg12:
If script is started through auto_rx.service file enabled in systemd, it fails due to lack of network connection with following output:

`Nov 10 11:20:45 alarm systemd[1]: Started auto_rx.
Nov 10 11:20:58 alarm auto_rx[220]: 2021-11-10 11:20:58,828 INFO:Reading configuration file...
Nov 10 11:20:58 alarm auto_rx[220]: 2021-11-10 11:20:58,875 WARNING:Config - APRS Update Rate clipped to minimum of 30 seconds. Please be respectful of other users of APRS-IS.
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,275 INFO:Config - Tested SDR #0 OK
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,307 INFO:Started Flask server on http://0.0.0.0:5000
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,315 INFO:Telemetry Logger - Started Telemetry Logger Thread.
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,338 ERROR:APRS-IS - Connection to APRS-IS Failed - [Errno -2] Name or service not known
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,347 INFO:APRS-IS - APRS Uploader Started.
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,351 INFO:OziMux - Started OziMux / Payload Summary Exporter
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,357 INFO:Sondehub Uploader - Started Sondehub Uploader Thread.
Nov 10 11:21:02 alarm auto_rx[220]: 2021-11-10 11:21:02,455 ERROR:Sondehub Uploader - Upload Failed: HTTPSConnectionPool(host='api.v2.sondehub.org', port=443): Max retries exceeded with url: /listeners (Caused by NewConnectionError('<u>
Nov 10 11:21:02 alarm auto_rx[220]: --- Logging error ---
Nov 10 11:21:02 alarm auto_rx[220]: Traceback (most recent call last):
[...]

PR #586, opened 2021-11-13
wait for network to allow proper start of script

dfgg12:
fix for #585


darksidelemm:
This needs to be checked to ensure it doesn't inhibit startup on 'standalone' (no internet or network connection) machines.
I don't think this will really solve the issue, which is more about a logging.exception line spamming the log with a huge traceback. auto_rx should continue on past this exception anyway.

PR #599, opened 2021-12-31
Custom nearby landing email subject

trgosk:
Useful for those who own more stations, they can add station name to nearby_landing_subject field in station.cfg

### DIFF
--- a/auto_rx/autorx/config.py
+++ b/auto_rx/autorx/config.py
@@ -206,7 +206,6 @@ def read_auto_rx_config(filename, no_sdr_test=False):
                 auto_rx_config["email_from"] = config.get("email", "from")
                 auto_rx_config["email_to"] = config.get("email", "to")
                 auto_rx_config["email_subject"] = config.get("email", "subject")
-                auto_rx_config["email_nearby_landing_subject"] = config.get("email", "nearby_landing_subject")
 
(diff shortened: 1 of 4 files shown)


darksidelemm:
Nice addition!

Could you move the line in config.py that reads the subject line into a new try/except section down here: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/autorx/config.py#L645

That way the added configuration setting is fully backwards compatible, so that stations that update automatically (there are a few out there now) don't see their station break.

darksidelemm:
Sorry, i haven't had time to look at this one, I'll try and get to this on the weekend.

auto_rx/autorx/config.py
```diff
@@ -205,6 +206,7 @@ def read_auto_rx_config(filename, no_sdr_test=False):
                 auto_rx_config["email_from"] = config.get("email", "from")
                 auto_rx_config["email_to"] = config.get("email", "to")
                 auto_rx_config["email_subject"] = config.get("email", "subject")
+                auto_rx_config["email_nearby_landing_subject"] = config.get("email", "nearby_landing_subject")
```
darksidelemm, line 209: I think this should be removed, as it is handled in the try/except lower down?

trgosk: Yes, this can be removed.

PR #610, opened 2022-02-07
[systemd service] use of environment variables

f4iey:
Hello there!
Is it possible to replace the annoying `/home/pi` by something like `$HOME` or `/home/$(whoami)` so that it matches any user home folder?

### DIFF
--- a/auto_rx/auto_rx.service
+++ b/auto_rx/auto_rx.service
@@ -3,10 +3,10 @@ Description=auto_rx
 After=syslog.target
 
 [Service]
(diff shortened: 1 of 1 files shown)


darksidelemm:
Have you tested that the systemd service with $HOME works? I'm guessing it probably will.

f4iey:
I just checked: it works if the service is enabled by the right user. 
But according to the wiki, `sudo` is used for starting/stopping the systemd service, so it points to the `/root` directory.
I guess the best thing to avoid using nano or any text editor (e.g during a packaged or automated install) is to add something like
```sh
sed -i "s/pi/${whoami}/g" auto_rx.service
```
somewhere in the `build.sh`
(Of course, admitting the user has a home directory)

PR #611, opened 2022-02-07
add sed for systemd service

f4iey:
Added sed so that the username is *automatically set* in `auto_rx.service`
Admitting the script is run by the right user (not as root)
I just tested and it worked fine for me

### DIFF
--- a/auto_rx/build.sh
+++ b/auto_rx/build.sh
@@ -82,4 +82,7 @@ cp ../demod/mod/meisei100mod .
 cp ../demod/mod/imet54mod .
 cp ../demod/mod/mp3h1mod .
 
+# Match user name in systemd service: no more need to use a text editor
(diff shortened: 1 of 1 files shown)


joyel24:
It should be merged in my opinion !

auto_rx/build.sh
```diff
@@ -82,4 +82,7 @@ cp ../demod/mod/meisei100mod .
 cp ../demod/mod/imet54mod .
 cp ../demod/mod/mp3h1mod .
 
+# Match user name in systemd service: no more need to use a text editor
+sed -i "s/pi/$(whoami)/g" auto_rx.service
```
pinkavaj, line 86: IMHO the Specifiers should be used here `%h` specifically

https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Specifiers

to make it more generic

Issue #619, opened 2022-02-28
APRS-IS duplicate packet filtering broken by sondegate inserting unique data


hessu:
The APRS-IS servers are capable of automatically filtering duplicate copies of the same packet, when the same packet is received by multiple receivers / gateways (igates).

This only works if the packet is forwarded intact, and the receiver does not insert unique data in the packet.

This only works correctly within a 30-second time window.  This means that packets must not be delayed - they must be forwarded immediately, not delayed for up to 30 seconds.

The sondegate software inserts data in the packet (timestamp, software version), data which is not sent by the sonde, but rather data that is specific to the gateway. This breaks duplicate packet filtering, and when multiple sondegates hear the same packet, it will be forwarded many times over the APRS-IS. This adds congestion on both APRS-IS and at services getting data from it.

Sondegate needs to be fixed so that when a single packet sent by a sonde is sent to APRS-IS, the following fields are exactly the same:
* The source callsign (unique to the sonde)
* The destination callsign (same for all sondes / sonde gateways - not specific to gateway software)
* The packet contents, beginning from the ':' all the way to the end of the line

jtmnt:
For starters, maybe remove auto_rx version number from description [https://github.com/jtmnt/radiosonde_auto_rx/commit/d58fec1f78981e0b71f4dcd15dfffea0aaa5ac36] (link)

darksidelemm:
This will work for auto_rx stations, but there are many other stations uploading to APRS, and they don't all do the same thing. In fact, many user go customize what is uploaded..

Note that the timestamp *is* from the radiosonde telemetry.

hessu:
It'd be great if there would be one gateway which works right, which can then be used as a reference implementation. I can then filter out the badly working / customized variants here, and point their authors to do the Right Thing.

If the timestamp within the packet is truely coming from the radiosonde telemetry, and not derived in any way from the local clock of the gateway, that's great! If all gateways forward it the same way, the packet will look the same and duplicates & delayed packets can be suppressed.

One thing that needs to be fixed is the APRS destination callsign. Now it seems to be allocated per gateway software, which will result in different variants of the same original packet. There should be just one dst callsign for all sondes, and a per-gateway-software tocall which can be used for the beacon indicating the location of the gateway itself, to convey its software variant.

TheSkorm:
hmmm. Would would the recommend dst call me?

TheSkorm:
[...]

Issue #618, opened 2022-02-28
Too frequent transmit rate, synchronised to create a DOS attack at APRS-IS services


hessu:
Hi,

The sondegate software appears to send packets at a **very high** rate to the APRS-IS. For example, here a single sonde is beaconing every 3 seconds **on average**, and this madness is amplified if there is more than one sondegate receiving that sonde at the same time:

https://aprs.fi/?c=raw&call=T1310381

Once a  minute on average should do fine, it seems to be alright for mostly everyone else on the APRS-IS and sondes should not receive special high-resolution treatment. If the rate is not reduced, I need to start filtering sondes off from aprs.fi to reduce load at this end.

Furthermore, it appears that the packets are sent every 30 seconds, synchronised to wall clock. This behaviour is explained in 
https://github.com/projecthorus/radiosonde_auto_rx/issues/373#issuecomment-769724610 . When all sondegates do this, there are two large synchronised bursts of packets every minute precisely at 12:00:00, 12:00:30, 12:01:00, 12:01:30, etc. The gateways are typically NTP synchronised, so they all do this blasting in unison. This fills up buffers on APRS-IS and clients receiving the data.

This needs to be be changed to forward packets immediately, and then stop forwarding further packets for 30 to 60 seconds for the same sonde.  The rate limit must not be synchronized to wall clock.

Packets must not be buffered and sent later - the APRS-IS duplicate filtering only works within a 30-second window.

73!

- Hessu, OH7LZB, aprs.fi

jtmnt:
> Packets must not be buffered and sent later

As I understand, buffering is there only to send one single latest packet after some time period. RS41 sonde sends different data in every 2 seconds.

TheSkorm:
3 seconds for a single station is certainly someone messing with the code. The config will usually clamp upload rate to 30 seconds : https://github.com/projecthorus/radiosonde_auto_rx/blob/a85a8d2fbe64ffb7d31d63bfc339d1d96b3546a8/auto_rx/autorx/config.py#L353

so this user has gone out of their way to have a higher upload rate it seems.

PR I put in moves the min upload rate up to 60 seconds

darksidelemm:
Hi @hessu , sorry about all this.

I wrote the synchronised upload part a long time ago, back when radiosonde_auto_rx only had a few users. Now there's ~600 users, it's clearly causing issues. 

[...]

PR #620, opened 2022-02-28
Fix/aprs performance

TheSkorm:
Changes min APRS time to 60 seconds. Removes version from APRS comment. Removes wall clock.

image

Does this look right to you @hessu ? 

Closes: 
- #619
- #618

### DIFF
--- a/auto_rx/autorx/aprs.py
+++ b/auto_rx/autorx/aprs.py
@@ -89,8 +89,8 @@ def telemetry_to_aprs_position(
 
     # TODO: RS41 Burst Timer
 
-    # Add on auto_rx version
-    _aprs_comment += " auto_rx v" + auto_rx_version
+    # Add on auto_rx at the end
+    _aprs_comment += " auto_rx"
 
     # Convert float latitude to APRS format (DDMM.MM)
     lat = float(sonde_data["lat"])
@@ -308,7 +308,7 @@ class APRSUploader(object):
         station_beacon_position=(0.0, 0.0, 0.0),
         station_beacon_comment="radiosonde_auto_rx SondeGate v<version>",
         station_beacon_icon="/r",
-        synchronous_upload_time=30,
+        upload_time=60,
         callsign_validity_threshold=5,
         upload_queue_size=16,
         upload_timeout=5,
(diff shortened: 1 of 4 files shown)


auto_rx/autorx/aprs.py
```diff
@@ -653,7 +652,7 @@ def upload_timer(self):
         """ Add packets to the aprs upload queue if it is time for us to upload. """
 
         while self.timer_thread_running:
-            if int(time.time()) % self.synchronous_upload_time == 0:
+            if datetime.datetime.now() > self.next_upload:
```
hessu, line 655: This is not much better, it now delays positions for up to 60 seconds, and they will arrive badly delayed. Other receivers have forwarded the same positions earlier.

A position that is received, must be forwarded _immediately_, and then additional positions from the same station should be suppressed for 30 seconds.

auto_rx/autorx/aprs.py
```diff
@@ -362,7 +360,8 @@ def __init__(
         self.aprsis_reconnect = aprsis_reconnect
         self.upload_timeout = upload_timeout
         self.upload_queue_size = upload_queue_size
-        self.synchronous_upload_time = synchronous_upload_time
+        self.upload_time = upload_time
+        self.next_upload = datetime.datetime.now() + datetime.timedelta(seconds=upload_time)
```
hessu, line 364: Using datetime.datetime.now() uses wall-clock time. When clock jumps, this code is confused. For example, a Raspberry Pi does not have a real-time clock, software starts up, then NTP synchronizes and the clock jumps.  It's better to use a monotonic clock which is not affected by wall-clock changes (Python 3 required, but what doesn't these days):

```
self.next_upload = time.monotonic() + upload_time

...

if time.monotonic() > self.next_upload:
```

auto_rx/autorx/aprs.py
```diff
@@ -89,8 +89,8 @@ def telemetry_to_aprs_position(
 
     # TODO: RS41 Burst Timer
 
-    # Add on auto_rx version
-    _aprs_comment += " auto_rx v" + auto_rx_version
+    # Add on auto_rx at the end
+    _aprs_comment += " auto_rx"
```
hessu, line 93: I believe this will still be different depending on software variant/version. The comment should only contain things from the sonde, not the gateway software, so that each gateway would send the exact same comment string.

auto_rx/autorx/aprs.py
TheSkorm: 60 seconds would be the max delay. It'll be uploading the latest sonde frame at that time. This is used to so we always report the last sonde frame. It's a bit hacky but typically the frame will be less than 1-2 seconds old.

auto_rx/autorx/aprs.py
TheSkorm: I'll fix this up, but note that autorx will still be sending to a different destination 
`      APRARX  VK5QI's radiosonde tracking` so I'm not sure if this will be that helpful for deduplication ?

auto_rx/autorx/aprs.py
TheSkorm: good catch :)

auto_rx/autorx/aprs.py
hessu: That will need to be fixed as well. We'll have to assign a common destination callsign for sondes at https://github.com/aprsorg/aprs-deviceid and use it instead of a gateway-specific destination call.

PR #632, opened 2022-04-25
Add a task preference item to each SDR

pamribeirox:
The default behavior is kept if station.cfg isn't updated
By default the SDR's are tasked simply from the lowest to the higher id (serial number)
Sometimes due to different RTL hardware, antennas or RF filtering/amplification is convenient to sort the task allocation based on a defined order.
This changes add an optional parameter to each SDR that can be used to raise or lower his allocation preference.

Hope this will help other people from the community

Python code changes by my son Diogo (I'm a C/C++ guy)

Regards CT7ABP Pedro.

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -92,6 +92,20 @@ gpsd_adaptor = None
 # This contains frequncies that should be blocked for a short amount of time.
 temporary_block_list = {}
 
+def get_sdr_item_pref(sdr_item):
+    """A helper function for the purpose of sorting the list of SDR by preference.
+
+    Args:
+        sdr_item (tuple) : Single element resulting from sdr_list.items(). A pair of ID (str) and config (dict).
+
+    Returns:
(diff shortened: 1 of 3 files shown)


darksidelemm:
Hi, thanks for working on this, it does seem like a useful feature.

First up, this PR *must* be made against the testing branch, not master. There are ongoing changes in testing.

Also, I am in the middle of a very big re-work of the entire SDR backend system at the moment which has resulted in some large changes to the codebase. This is unfortunately not yet in the testing branch on this repository as it is still very much in-work, but it will possibly result in conflicts with this PR.

As such, I'll probably need to hold off on merging in this PR until these other changes are done.

darksidelemm:
Hi, I've now merged my the bulk of the SDR wrapper changes into the testing branch. If you could please re-target this PR to that branch that would be appreciated.

Note that in the networked-SDR mode (which lets auto_rx connect into a SpyServer), there should be no preference given to any particular 'virtual' SDR.

pamribeirox:
Hello,

I'll do that as soon as possible it will probably take some days as I'm 
in a work peak.
regards.

PR #640, opened 2022-07-11
Testing - Add a task preference item to each SDR (ignored the new networked receivers, just port from 1.5)

pamribeirox:
The default behavior is kept if station.cfg isn't updated
By default the SDR's are tasked simply from the lowest to the higher id (serial number)
Sometimes due to different RTL hardware, antennas or RF filtering/amplification is convenient to sort the task allocation based on a defined order.
This changes add an optional parameter to each SDR that can be used to raise or lower his allocation preference.

Hope this will help other people from the community

Python code changes mostly by my son Diogo (I'm a C/C++ guy)

Regards CT7ABP Pedro.

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -93,6 +93,17 @@ gpsd_adaptor = None
 # This contains frequncies that should be blocked for a short amount of time.
 temporary_block_list = {}
 
+def get_sdr_item_pref(sdr_item):
+    """A helper function for the purpose of sorting the list of SDR by preference.
+    Args:
+        sdr_item (tuple) : Single element resulting from sdr_list.items(). A pair of ID (str) and config (dict).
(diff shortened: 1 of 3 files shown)


darksidelemm:
I think this will break if the networked-SDR configuration is used, as the code will be looking for a 'preference' field.

You'd need to add in a preference field when the network SDR structures are created:
https://github.com/projecthorus/radiosonde_auto_rx/blob/47246b73c6bdade8f7422162cc03895d2a1e2a03/auto_rx/autorx/config.py#L791
and
https://github.com/projecthorus/radiosonde_auto_rx/blob/47246b73c6bdade8f7422162cc03895d2a1e2a03/auto_rx/autorx/config.py#L813

darksidelemm:
Sorry, i think this one is way out of date now.

PR #649, opened 2022-08-04
Minor changes to the display style of the log bars and settings

Piter-NS:
The changes concern the size of the sliders, the font and the arrangement of elements. By the way, I changed the way of displaying the date and time for the UTC format

### DIFF
--- a/auto_rx/autorx/static/css/main.css
+++ b/auto_rx/autorx/static/css/main.css
@@ -52,12 +52,11 @@ a {
   height: 100px;
   width: 100%;
   position: absolute;
-  z-index: 10;
+  z-index: 100;
   text-align: center;
   bottom: 0;
   left: 0;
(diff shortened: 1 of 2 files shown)


darksidelemm:
Can you please provide some before/after screenshots of what has changed here?

If possible, i would prefer it if the local/UTC switch was an option, not something fixed. Some prefer to see local time, others prefer UTC time.

Piter-NS:
Before changes
log_bar_before settings_bar_before

After changes
log_bar_after settings_bar_after

The local / UTC switch is still an option. I just changed the way UTC is displayed from 2022-08-04T09:54:39Z to 2022-08-04 09:54:39 UTC.

darksidelemm:
OK, thanks for this - we'll see how the users on the testing branch find it.

PR #662, opened 2022-09-20
Increase the crop for rtl_power 20% -> 25%

pinkavaj:
Some SDRs have need this to avoid significat vallyes in the spectrum of the scan


pinkavaj:
This should slightly improve the detection of the signal.

PR #664, opened 2022-09-20
Replace build commands in build.sh by running a make

pinkavaj:


darksidelemm:
Thanks for these PRs! These do a lot of much-needed cleanup of this repository... 
I need some time to go through and review them, since some have a *lot* of changes (in particular the pre-commit one, which I think i might prefer to run locally rather than have a PR with every single file changed).

pinkavaj:
Sure, btw this PR replaces `-Ofast` by `-O3`  for some programs, not sure if this is or is not an issue.  Is it?

darksidelemm:
I'd try and keep them the way they were if possible. -Ofast did give some speedups that were helpful on older RPis, though we didn't use it everywhere as some of the decoders behaved oddly with -Ofast

pinkavaj:
I have fixed it, the `-Owhatever` behaviour should be the same as in old `build.sh`.

darksidelemm:
Thanks again for this! This makes the whole process a lot cleaner.

I have made some modifications to inhibit some of the warnings. Not best practice I know, but I also don't really have time to fix them all, and the flood of warnings will probably give some new users concerns :-)

pinkavaj:
> Thanks again for this! This makes the whole process a lot cleaner.
> 
> I have made some modifications to inhibit some of the warnings. Not best practice I know, but I also don't really have time to fix them all, and the flood of warnings will probably give some new users concerns :-)

I can help to fix some of them. But I'm not sure how far is acceptable to go, because it might cause lot of changes in the codebase ...

Btw #668 was result of goinng throught warnings using `clang` as a compiler.

darksidelemm:
All of the decoders are written by @rs1729, not me. It might be better to make PRs to their repository ( https://github.com/rs1729/RS ) rather than this one.

Issue #6, opened 2017-12-16
Move RS to a submodule


TheSkorm:
This project should focus specifically on auto_rx and just use a submodule version of RS. This will need to a seperate repo with the JSON patches included and instructions updated on how to pull down submodule.

darksidelemm:
I think this is probably not necessary anymore. If anything we'll move the RS code into a src subdirectory.

PR #668, opened 2022-09-24
Fix memset parametrization

pinkavaj:
The sizeof() and constant part were swapped.

### DIFF
--- a/demod/mod/dfm09mod.c
+++ b/demod/mod/dfm09mod.c
@@ -1707,7 +1707,7 @@ int main(int argc, char **argv) {
         int pos = 0;
         float _frmcnt = -1.0f;
 
-        memset(buffer_rawhex, BUFLEN+1, 0);
+        memset(buffer_rawhex, 0, BUFLEN+1);
 
         while ( (ch=fgetc(fp)) != EOF)
         {
--- a/demod/mod/m10mod.c
+++ b/demod/mod/m10mod.c
@@ -1522,7 +1522,7 @@ int main(int argc, char **argv) {
 
         while (1 > 0) {
 
-            memset(buffer_rawhex, 2*(FRAME_LEN+AUX_LEN)+12, 0);
+            memset(buffer_rawhex, 0, 2*(FRAME_LEN+AUX_LEN)+12);
             pbuf = fgets(buffer_rawhex, 2*(FRAME_LEN+AUX_LEN)+12, fp);
             if (pbuf == NULL) break;
             buffer_rawhex[2*(FRAME_LEN+AUX_LEN)] = '\0';
--- a/demod/mod/m20mod.c
+++ b/demod/mod/m20mod.c
@@ -1355,7 +1355,7 @@ int main(int argc, char **argv) {
 
         while (1 > 0) {
 
-            memset(buffer_rawhex, 2*(FRAME_LEN+AUX_LEN)+12, 0);
+            memset(buffer_rawhex, 0, 2*(FRAME_LEN+AUX_LEN)+12);
             pbuf = fgets(buffer_rawhex, 2*(FRAME_LEN+AUX_LEN)+12, fp);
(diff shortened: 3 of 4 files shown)


rs1729:
@pinkavaj 
Thanks for spotting this. Would you make a pull request at
https://github.com/rs1729/RS/tree/master/demod/mod
(m20mod.c aka mXXmod.c)
Them files more or less:
```
dfm09mod.c:1710:        memset(buffer_rawhex, BUFLEN+1, 0);
m10mod.c:1525:            memset(buffer_rawhex, 2*(FRAME_LEN+AUX_LEN)+12, 0);
//  meisei100mod.c:355:    dsp_t dsp = {0};  //memset(&dsp, 0, sizeof(dsp));
meisei100mod.c:705:                            memset(&gpx, sizeof(gpx), 0);
meisei100mod.c:870:                            memset(&gpx, sizeof(gpx), 0);
mXXmod.c:1363:            memset(buffer_rawhex, 2*(FRAME_LEN+AUX_LEN)+12, 0);
```

pinkavaj:
> @pinkavaj Thanks for spotting this. Would you make a pull request at https://github.com/rs1729/RS/tree/master/demod/mod (m20mod.c aka mXXmod.c) Same files more or less:

Sure, I'm actually now going throught both repositories and just thinking if it would not be worth to sync them and eg. use the RS repository just as a git submodule here. It will deduplicate the code, both sides will get fresh updates and the attention would not be spread accross multiple repositories.

Is this somethink which can be considered as an option?

rs1729:
Well, I suggested a submodule some time ago. I was thinking of making a new repository for that with only the stuff used in auto_rx, since in the RS-repo there are many things for historical reasons etc., it is more like a notebook I like to say.

rs1729:
@darksidelemm 
fixed upstream.

pinkavaj:
> Well, I suggested a submodule some time ago. I was thinking of making a new repository for that with only the stuff used in auto_rx, since in the RS-repo there are many things for historical reasons etc., it is more like a messy notebook I like to say.

I see the repository would need a bit of clean-up, but from what I see people are used to it. Sinc this is git, would it make sense to really just try to polish a bit the RS repository and re-consider the submodule stuff. I'm not yet too deep in this, but my experience is, forking is easy, but joining anythink back is the hard part ...

rs1729:
I would like to have a cleaner/stripped `RS_autorx` repository for auto_rx that could be used as a submodule or in a subdirectory. Sometimes I'm not happy to make changes to the RS-repo even if it would make the code cleaner or more usable in auto_rx.
I would reconsider
https://github.com/projecthorus/radiosonde_auto_rx/issues/6
@darksidelemm ?

darksidelemm:
Given we do occasionally make auto_rx specific changes to things, I'm still fine with just bringing in the updates as required, instead of going with a submodule.

rs1729:
meisei100mod.c:

After fixing the memset() call, I realized that a complete reset of `gpx` that I had in mind would affect decoding (if a reset occurred).
`frame_(raw)bits` and Reed-Solomon `RS` must not be reset.

Fixed up stream.
```
-                        if (reset_gpx) {
-                            memset(&gpx, 0, sizeof(gpx));
+                        if (rst_gpx) {
+                            reset_gpx(&gpx);
```

Remark on `-Ofast/--fast-math`:
In `demod_mod.c` it speeds up floating point calculations, inaccuracies should be negligible, didn't observe unusual behavior.
However `meisei100mod.c` shouldn't be compiled with the fast-math flags because it is using the `isnan()` function, it would e.g. output `nan` values in JSON if temperature/humidity is nan.
I would (and do) use `-Ofast` only in `demod_mod.c` (and `dft_detect.c`, `imet4iq.c`, `mk2a1680mod.c`), `-O3` otherwise.

darksidelemm:
I think we can close this one now, I've included the changes from upstream.

Issue #628, opened 2022-03-14
Multi-Channel Decode Thoughts


darksidelemm:
Trying to collate some thoughts on what auto_rx would need from a multi-channel SDR server (think: SpyServer, but actually open source) into one place, instead of scattered across multiple issues.

auto_rx needs two things to be able to receive radiosondes:

### RF Spectrum Snapshot / Average
auto_rx needs a power spectral density snapshot of the RF spectrum over the frequency range of interest. auto_rx performs peak detection on this (with a user-configurable SNR threshold) to decide what frequencies it should try and receive radiosonde signals on.

This doesn't necessarily need to be power spectral density - it could also be an SNR estimate.

It important that this snapshot of spectrum is actually an average (or possibly a max hold?) over a definable time period. Many radiosonde types do not transmit continuously, so we need to be observing for at least 1 second to ensure we catch those sonde types.

Currently, auto_rx uses rtl_power to get this information. rtl_power will perform power averaging over a user-definable time period (auto_rx sets this to 20 seconds by default), and provide a CSV output. Currently we use a frequency bin size of 800 Hz. Another neat thing rtl_power will do is jump the SDR in frequency to cover a frequency range wider than the SDR's bandwidth. Not required if you have a SDR that will cover the entire range of interest of course...

### IQ / FM-demodulated RF
Once we have a candidate radiosonde frequency, auto_rx then performs a 'detection' step. It feeds 48 kHz BW IQ (signed 16-bit) into the 'dft_detect' utility, which performs correlation against known radiosonde packet headers to determine if the candidate signal is in fact a radiosonde, and which type it is. It will also estimate the actual centre frequency of the radiosonde. This information is then returned to auto_rx.

auto_rx then starts up the appropriate decoder. The majority of radiosonde decoders in auto_rx take IQ input (signed 16-bit format). Due to some optimisations in the FSK demodulator used (fsk_demod), the IQ bandwidth must be an integer multiple of the signals baud rate.  Currently we use rtl_fm to provide this narrowband IQ (`-M raw`) The following sample rates are used:
* RS41, RS92, M20, LMS6, iMet-54, MRZ: 48000 Hz (2400/4800/9600 baud)
* RS92-NGP: 96000 Hz (still 4800 baud, but higher modulation index)
* DFM: 50000 Hz (2500 baud)
* M10: 48080 Hz (9616 baud)

[...]

PR #676, opened 2022-09-24
Allow use different types of receivers

pinkavaj:
Allow specify for each device type what kind of SDR it is, this allow combining networked and local SDRs from various sources or just use multiple network devices.

### DIFF
--- a/auto_rx/autorx/config.py
+++ b/auto_rx/autorx/config.py
@@ -284,7 +284,7 @@ def read_auto_rx_config(filename, no_sdr_test=False):
         if auto_rx_config["station_lat"] > 90.0 or auto_rx_config["station_lat"] < -90.0:
             logging.critical("Config - Invalid Station Latitude! (Outside +/- 90 degrees!)")
             return None
-        
+
         if auto_rx_config["station_lon"] > 180.0 or auto_rx_config["station_lon"] < -180.0:
             logging.critical("Config - Invalid Station Longitude! (Outside +/- 180 degrees!)")
             return None
@@ -632,7 +632,7 @@ def read_auto_rx_config(filename, no_sdr_test=False):
             )
             auto_rx_config["web_control"] = False
             auto_rx_config["web_password"] = "none"
-        
+
         try:
             auto_rx_config["save_raw_hex"] = config.getboolean(
                 "debugging", "save_raw_hex"
@@ -642,7 +642,7 @@ def read_auto_rx_config(filename, no_sdr_test=False):
                 "Config - Did not find save_raw_hex setting, using default (disabled)"
             )
             auto_rx_config["save_raw_hex"] = False
-        
+
         try:
             auto_rx_config["experimental_decoders"]["MK2LMS"] = config.getboolean(
                 "advanced", "lms6-1680_experimental"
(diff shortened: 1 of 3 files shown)


pinkavaj:
@rs1729 Hi, just got this idea and investigating if it is right time to go this way. The code is not finished nor properly tested, the goal was just to show the idea. Best to see changes in sample config to get whats going on.

darksidelemm:
I would prefer not to make this change. 

When in 'network SDR' mode, auto_rx assumes the network sdr supports multiple demodulators, and since we only support spyserver, that is definitely the case. The 'number of sdrs' option sets the number of demodulators to be used in parallel, which is mainly going to be limited by CPU.

When in RTLSDR mode, it does require one RTLSDR per demodulator.

I don't want to mix these two modes of operation.

darksidelemm:
If you're on discord, it might be worth jumping on the #radiosonde_auto_Rx channel, where I, and a few other devs hang out to discuss: https://discord.gg/8sWUQKGY

pinkavaj:
> If you're on discord, it might be worth jumping on the #radiosonde_auto_Rx channel, where I, and a few other devs hang out to discuss: https://discord.gg/8sWUQKGY

Unfortunately not now.

pinkavaj:
@darksidelemm Hi, not sure what the discussion went, but from my point of view it is even a bit of waste to allow receive only 1 sonde on rtl-sdr, when there is the ~2.4MHz BW and in most cases the 8bit resolution is enough to receive many. But of course nontrivial amont of would would be needed to do this.

darksidelemm:
Sure, I agree that the current way we use RTLSDRs is wasteful, however I do not want to have the DSP required to do this kind of thing (e.g. on-demand channelisation) within the auto_rx repository.

I have already described in detail what auto_rx needs in terms of spectrum and IQ data to operate here: https://github.com/projecthorus/radiosonde_auto_rx/issues/628

In the current testing branch, I have included the ability to connect to an SpyServer instance, which can do all of the above. With an Airspy Mini or Airspy R2, essentially the entire radiosonde band can be monitored simultaneously. 
RTLSDRs can also be used with SpyServer, but support for them appears to have been somewhat hobbled by the developer (who wants to sell Airspy units...).

The SDR wrapper code that is now in auto_rx should be fairly easily extended to support other spectrum/IQ sources when they become available. KA9Q's server is one that I hope to add support for, and hopefully some other options present themselves.

pinkavaj:
Ok, lets close this. I will try to align with project goals and strategy. The KA9Q stuff looks interesting, i will do at least some exploration related to this problem here.

darksidelemm:
I believe @rs1729 also has a server that he has been working on that might work too.

rs1729:
Probably it would be easier to have it all integrated in one program and restrict to certain configurations, but that would be a different approach. In the modular approach of auto_rx you need some kind of server/clients communication.

There is
https://github.com/rs1729/RS/tree/master/demod/multi
that can decode certain types of radiosondes simultaneously. It has a basic way of adding and removing signals, but you would also need to scan the frequency band from time to time.

Don't know if it still possible, but you could use it manually via the UDP option in auto_rx.

https://github.com/rs1729/RS/tree/master/demod/iq_svcl
(https://github.com/projecthorus/radiosonde_auto_rx/issues/628#issuecomment-1073221184)
could provide IF IQ data for the decoders in a more general way, but I never tried to integrate it into auto_rx or other kind of automated monitoring and decoding.
And I don't know how it would perform on a RPi 4, probably not really good. But that's ok, these are only ideas of how it could work in principle, performance was not the primary goal.

darksidelemm:
I think the IQ server/client approach is the right way about it, for auto_rx integration.

rs1729:
However to perform on RPi4, it may be good to have a modified version that restricts to certain baseband and IF sample rates. This could be easier optimized (in particular the decimation).
Though I guess KA9Q's server will have good performance, maybe that will be the right alternative for auto_rx.

PR #692, opened 2022-10-26
Clean-up, remove unnecessary spaces, newlines and other text noise

pinkavaj:
Most modern text editors remove them automatically and even if not it is a bit inconvenient because new whitespaces are sometimes added or removed during code editing by accident and needs to be removed manual to send pull-requests without this noise.

PS: Feel free to run `pre-commit run -a` manually your-self. I guess it is impossible to actually review this.
PPS: Would it actually make more sense to run this once the new release is out and the `testing` is in sync with `master`?


darksidelemm:
Yes, as mentioned in a previous PR, I'd prefer to avoid merging such a widespread change.

pinkavaj:
> Yes, as mentioned in a previous PR, I'd prefer to avoid merging such a widespread change.

Sure I would neither. Sorry for my impatience, It just triggers my OCD :). This is intended just as a remainder not to forget it, but NP to close it, I does not need to be obnoxious.

PR #698, opened 2022-11-12
Add dft_detect support for MTS01, drop support for IMET1AB

darksidelemm:


icalik:
@darksidelemm when it will be in testing branch? I can pull and test at today's sounding.

darksidelemm:
I've merged this now. No guarantees on performance. I'm still using the fixed detection bandwidth to enable detection slightly off-frequency. Will need to think a bit more about that setting.

icalik:
I've set `save_decode_iq` and `save_raw_hex` configs as True, collected files will be used for performance testing.

icalik:
@darksidelemm 
I just tested it with MTS01 from home. dft_detect detect wrong freq on 2 iterations, 3rd one it found as 401MHz. 
And when trying to upload to sondehub, the response is as invalid, should we put MTS01 before serial?


```
{
            "error_message": "Payload ID 10010365 is invalid.",
            "payload": {
                "software_name": "radiosonde_auto_rx",
                "software_version": "1.6.0-beta16",
                "uploader_callsign": "TB2ISC-11",
                "uploader_position": [
                    40.9235,
                    29.3021,
                    126.0
                ],
                "uploader_antenna": "UHF Antenna",
                "time_received": "2022-11-12T09:41:10.889784Z",
                "datetime": "2022-11-12T09:41:09.000000Z",
                "manufacturer": "Meteosis",
                "type": "MTS01",
                "serial": "10010365",
                "frame": 736,
                "lat": 40.92315,
                "lon": 29.30213,
                "alt": 125.0,
                "frequency": 401.001,
                "ref_position": "MSL",
                "ref_datetime": "UTC",
                "user-agent": "Amazon CloudFront",
                "position": "40.92315,29.30213"
            }
        }
```

darksidelemm:
There should be no need to do that. We need to make some changes on the sondehub side for it to work correctly. 

Not sure what you mean by detected wrong frequency?

darksidelemm:
On the point of the serial numbers getting into sondehub - the MTS01 serial beginning with S should have never actually made it onto the map. However, I had an error in a regex check on the serial numbers, so it got through. 

We've made the change now on sondehub so the serials from the latest auto_rx testing should get through OK.

PR #716, opened 2022-11-18
Update send_file call to avoid deprecated argument

darksidelemm:


argilo:
This change breaks log file downloads on "native" installs, because Raspberry Pi OS (Bullseye) ships with Flask 1.1.2. I guess the problem will be mostly resolved once Rasperry Pi OS updates to Bookworm (which ships with Flask 2.2.2).

darksidelemm:
So if you remove the python3-flask package, then install flask via pip (e.g. pip3 install -r requirements.txt ) that should bring in the latest flask. 
I've removed python3-flask from the dependency install line in the install guide.

argilo:
> I've removed python3-flask from the dependency install line in the install guide.

That seems like a good solution.

Installing flask through pip did solve the problem, as expected.

darksidelemm:
Good to hear! Apologies about this, I haven't been focusing as much on the 'native' install instructions recently.

PR #725, opened 2022-11-23
Update mts01mod for hor.vel and temp

icalik:
I tested it with MTS01 probe and then I was able to see temperature value
Updated README file to enable temp on Meteosis MTS01

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ auto_rx/imet4iq
 auto_rx/imet54mod
 auto_rx/mXXmod
 auto_rx/mp3h1mod
+auto_rx/mts01mod
 
 m10
 meisei100mod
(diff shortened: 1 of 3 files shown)


darksidelemm:
@rs1729 have you seen this?

icalik:
> @rs1729 have you seen this?

yes i hve copied from there, like what you did. is it a problematic thing?

darksidelemm:
Nope, that's fine! I hadn't spotted the changes. Will merge in now.

rs1729:
Maybe my comment was not so clear
https://github.com/rs1729/RS/issues/48#issuecomment-1321264578
but that was about the latest mts01mod commit,
Temperature and horizontal velocity.
I didn't have information about the temperature sensor, though it looks like a thermistor. The approximation I chose didn't look too bad. I compared three soundings.
Could be that very high and very low temperatures show more deviation.
For relative humidity, my first impression was it needs calibration data. Or maybe the sensor/measurement was off at high altitude for one of the soundings, it showed a different offset than the other two. Plotting the raw data looked like there might be a simple linear mapping, but I didn't check yet. Maybe it depends on calibration data.

rs1729:
@darksidelemm 
No, I didn't see the pull request here. There was no reaction to my comment/commit upstream.
@icalik 
Next time you take a file from upstream, you could mentioned it in the pull request or commit message, then it is clear that it is the same file, same changes.

PR #731, opened 2022-12-11
SpyServer + AIRSPY mini does not work well in Japan

AstroTaka:
I started to use SpyServer + AIRSPY mini combination.
But it does not work in Japan sonde band.

In Japan, the sonde band is 403.3MHz-405.7MHz.
Center of the band is 404.5MHz.
I set 404.5MHz to spyserver.config.
I use 6Msps because 3Msps of SpyServer cannot cover the band.
And then, it can cover 402.1MHz-406.9MHz.

test_sdr in sdr_wrapper.py use fixed 401.5MHz.
auto_rx.py exit because my setting of SpyServer does not cover 401.5MHz.

So I made a patch which use the frequency come from min_freq of station.cfg
decode.py use test_sdr.
But it does not have min_freq.
So I used sonde_freq.

I still have a issue.
A scan list include outside sonde band because SpyServer cover wider than Japan sonde band.
So I made a patch to remove outside min_freq and max_freq.

Please check!

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -167,7 +167,6 @@ def start_scanner():
             rtl_fm_path=config["sdr_fm"],
             rtl_device_idx=_device_idx,
             gain=autorx.sdr_list[_device_idx]["gain"],
-            agc=autorx.sdr_list[_device_idx]["agc"],
             ppm=autorx.sdr_list[_device_idx]["ppm"],
             bias=autorx.sdr_list[_device_idx]["bias"],
             save_detection_audio=config["save_detection_audio"],
@@ -251,7 +250,6 @@ def start_decoder(freq, sonde_type, continuous=False):
             rtl_fm_path=config["sdr_fm"],
             rtl_device_idx=_device_idx,
             gain=autorx.sdr_list[_device_idx]["gain"],
-            agc=autorx.sdr_list[_device_idx]["agc"],
             ppm=autorx.sdr_list[_device_idx]["ppm"],
             bias=autorx.sdr_list[_device_idx]["bias"],
             # Other options
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -167,6 +167,7 @@ def start_scanner():
             rtl_fm_path=config["sdr_fm"],
             rtl_device_idx=_device_idx,
             gain=autorx.sdr_list[_device_idx]["gain"],
+            agc=autorx.sdr_list[_device_idx]["agc"],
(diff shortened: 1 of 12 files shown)


darksidelemm:
OK, so these fixes seem reasonable. Sorry, I didn't realise the test frequency was fixed. 

The scan results shouldn't return anything outside of the range the spyserver is set to, so I don't think the min/max removal stuff is really necessary.

It might be better to have the check frequency be in the middle of the min/max range?

AstroTaka:
About the min/max removal stuff, SpyServer does not return anything outside of SpyServer range(4.8MHz). but the range is wider than Japan sonde band(2.4MHz). and it return a outside of frequency of Japan sonde band(402.1-403.3, 405.7-406.9).
So I hope to remove 402.1-403.3, 405.7-406.9.

If possible, remove the outside frequency from the scan plot is better I think because each signal of scan plot of SpyServer is smaller than RTL-SDR.
https://drive.google.com/file/d/1Qq2GLFjJ0ZjeVRHOgsT5pukOHRDi7OKY/view?usp=share_link

About the check frequency,  a center of min_freq and max_freq is better I think too.

darksidelemm:
OK, I think I understand your issue a bit more now. Are you ending up with a lot of peaks outside the sonde band? 

Are you able to make the changes to use the centre of min/max when running the tests?

AstroTaka:
Yes, a lot of peaks outside the sonde band and the signal sometimes very strong.
So I hope to remove the outside band from the scan result frequency list.

About the center of the band frequency check, I modified the code and start to check it.
It seems to be OK.
After receive one sonde, I will inform you.

P.S. I merged and revert a not related branch because I am not familiar with git operation...

AstroTaka:
The center of the band frequency check works fine for me.

darksidelemm:
Could this:
```
check_freq = (self.min_freq + (self.max_freq-self.min_freq)/2.0) * 1e6
```
just be:
```
check_freq = 1e6*(self.min_freq + self.max_freq)/2.0
```
??

AstroTaka:
It just follow your? code

https://github.com/projecthorus/radiosonde_auto_rx/blob/bbc83e424188150372e26cf13937588426ddba55/auto_rx/autorx/sdr_wrappers.py#L556

This is no  problem and more simple I think.

`check_freq = 1e6*(self.min_freq + self.max_freq)/2.0`

darksidelemm:
Thanks for this! I'll give this a test out on my local spyserver system too.

AstroTaka:
Thanks!

PR #740, opened 2022-12-27
Support RTL-AGC of another fork of librtlsdr(rtl_power and rtl_fm)

AstroTaka:
There is a setting of RTL-AGC in ExtIO of RTL-SDR.
Original fork of librtlsdr(rtl_power/rtl_fm) does not seems to support RTL-AGC.
I found one of another fork of librtlsdr which support RTL-AGC.

https://www.rtl-sdr.com/new-updates-to-the-librtlsdr-rtl-sdr-driver-fork/
https://github.com/librtlsdr/librtlsdr

"-e agc" is a option to enable RTL-AGC with rtl_power and rtl_fm.
So I added a function to enable/disable RTL-AGC by [sdr_x] of station.cfg
Default is disabled.
Original fork of rtl_power and rtl_fm seems to ignore it because the original already support "-e" option.

https://github.com/osmocom/rtl-sdr

I compared between enable and disable of RTL-AGC with RTL-SDR v3 .
Enable RTL-AGC was better performance than disable in particular very weak signal situation (phase of beginning and ending).

I compared between AIRSPY mini and RTL-SDR v3 with enable RTL-AGC.
Basically, RTL-SDR v3 is between than AIRSPY mini in very weak signal situation.
But AIRSPY mini decode(much decode count) better than RTL-SDR when not so weak signal level.

Please check it.

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -167,7 +167,6 @@ def start_scanner():
             rtl_fm_path=config["sdr_fm"],
             rtl_device_idx=_device_idx,
             gain=autorx.sdr_list[_device_idx]["gain"],
-            agc=autorx.sdr_list[_device_idx]["agc"],
             ppm=autorx.sdr_list[_device_idx]["ppm"],
             bias=autorx.sdr_list[_device_idx]["bias"],
             save_detection_audio=config["save_detection_audio"],
@@ -251,7 +250,6 @@ def start_decoder(freq, sonde_type, continuous=False):
             rtl_fm_path=config["sdr_fm"],
             rtl_device_idx=_device_idx,
             gain=autorx.sdr_list[_device_idx]["gain"],
-            agc=autorx.sdr_list[_device_idx]["agc"],
             ppm=autorx.sdr_list[_device_idx]["ppm"],
             bias=autorx.sdr_list[_device_idx]["bias"],
             # Other options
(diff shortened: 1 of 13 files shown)


darksidelemm:
Hmm, rather than adding new options in the configuration file, I'm wondering if we can just overload the current gain value?
Currently if the gain value is set to -1, the 'normal' AGC is used. Perhaps if this is set to -2, enable the new AGC Option? 

I would prefer this rather than adding a new configuration option.

AstroTaka:
I need to explain more detail about RTL-AGC.

This is good explanation.
https://www.reddit.com/r/RTLSDR/comments/77t0j0/someone_can_explain_me_all_this_agc_control_in_sdr/

"gain=-1" is "Tuner AGC" of the above explanation.
RTL-AGC is a AGC of RF baseband chip.
Tuner gain and RF  baseband gain work in different chip.

So basically, it need to setup separately each other.
But in my use case, I set agc=-1 and enable RTL-AGC.
So if "agc=-2" means "agc=-1 and enable RTL-AGC" is OK for me.

darksidelemm:
Yes, I would prefer that behaviour rather than having a separate option.

AstroTaka:
OK, I try to modify.

darksidelemm:
OK, now I understand a bit more. So RTL-AGC is baseband gain in the RTL2832 IC itself... this is not something I'd want enabled all the time I think, I can see how it could result in distortion with strong signals. Not sure I like the idea of that.

Anyway, having it enabled with -2 is fine with me, so I think your current solution is fine.

AstroTaka:
Thank you very much! and good new year!

PR #742, opened 2023-01-03
Remove MEISEI sonde type from  ID

AstroTaka:
I use beta25.

When I receive MEISEI iMS-100, the ID of main screen contain 'IMS100-'.
So I modified it which follows DFM, M10, M20, IMET, IMET5, IMET54 and MRZ.

I have no issue in SondeList about ID of MEISEI.

Before this patch of main screen 
before_meisei_id_fix_patch

After this patch of main screen
after_meisei_id_fix_patch

### DIFF
--- a/auto_rx/autorx/templates/index.html
+++ b/auto_rx/autorx/templates/index.html
@@ -483,8 +483,8 @@
                     {title:'Freq (MHz)', field:"freq", headerSort:true},
                     {title:"ID", field:"id", width:125, headerSort:true, formatter:function(cell, formatterParams, onRendered){
                         _cell_data = cell.getData();
-                        _id = _cell_data.id.replace(/^(DFM|M10|M20|IMET|IMET5|IMET54|MRZ|IMS100|RS11G)-/,"");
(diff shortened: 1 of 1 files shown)


darksidelemm:
There's actually 2 replaces there, both need to be changed. Suggest changing them both to include:
```
DFM|M10|M20|IMET|IMET5|IMET54|MRZ|LMS6|IMS100|RS11G|MTS01
```

This will bring it into line with what's used on the python side.

AstroTaka:
Of cause I know.
Probably this line.
Is it correct? 

https://github.com/projecthorus/radiosonde_auto_rx/blob/d4d2f26b398fba49ede642e76d0f0b5788d143ac/auto_rx/autorx/templates/index.html#L487

I could not change it because it ls related with SondeHub I think.
And currently, I have no issue about ID at SondeHub and iMS-100 combination.

I ask you double confirm.
Can I change it?

darksidelemm:
Yes, that's the correct line. just below the one you modified.

AstroTaka:
OK, Thanks!

darksidelemm:
That change looks fine, but could you please add MTS01 as well, like in my example further up?

AstroTaka:
OK!

AstroTaka:
@darksidelemm please check the updated file.

darksidelemm:
Yep, all fine.

Issue #729, opened 2022-12-01
Wish / Nice to have : Sonde Frequency in Historical


zanco:
Hi, is it possible to add the sonde frequency in the Sonde List which appears when you view Historical Data ?

Sometimes I see a sonde on the waterfall, know the frequency but I cannot sort the Sonde List on frequency.

Thanks,
Ben

darksidelemm:
We're kind of running out of width in that list! What would be better is to have some settings to allow what fields are displayed, like on the main page. This is a lot more work though.

darksidelemm:
Resolved in 1.6.1

PR #756, opened 2023-02-27
Add a frequency column to the Sonde List and add a visibility setting of each column of it.

AstroTaka:
I hoped to add a frequency column to the Sonde List same as #729.
So I made it based on a suggestion of the following comment.
https://github.com/projecthorus/radiosonde_auto_rx/issues/729#issuecomment-1362472187

I added a settings icon of Sonde List to the same position as "main" page.
I have one issue of a icon position of "return to main" and "SkewT".
I think that it is better that a icon position of "return to main" and "go to history" is same as current version.
So I swapped them.

Please check!


darksidelemm:
This looks good! I'm going to make a few tweaks once it's in the testing branch, but otherwise it's a good addition!

AstroTaka:
@darksidelemm Thanks!

PR #768, opened 2023-04-30
fix little mistake in m20mod.c comments

joyel24:
No need to comment I think...


darksidelemm:
I'll let @rs1729 comment on the accuracy of this one, since it's their code...

joyel24:
> I'll let @rs1729 comment on the accuracy of this one, since it's their code...

4 bytes to encode speed really makes nonsense : 256^4=4 294 967 296 possible values...
This is a comment only, it does not affect the soft but for people who want to understand the packets sent by M20 radiosondes to decode with other stuff like hackrf/portapack in my case, it can be confusing.
Anyone can understand why this is a mistake, look :

foto_no_exif-6

joyel24:
Oh no, I am the mistake...
There are 4 bytes for vertical velocity : 2 bytes for X (North/South) - 2 bytes  for Y  (West/East)

PR #777, opened 2023-05-29
Prototype: Run readsb during scan downtime

pasky:
I would like my SDR to track radiosondes as a priority, but receive other interesting things during the downtime - most obviously, contribute ADSB data.

This PR isn't a serious proposal on how to implement that, obviously (though it does work as a minimal PoC...). I'd like to use it more as a discussion ground, whether this kind of functionality would be welcome by you.

Thoughts:

- Would a generalized configuration "pair of commands for post-scan hook, pre-scan hook" configuration be a viable approach, for example?
- I'd ideally like to have readsb just have an API to pause its SDR usage rather than starting/stopping it completely (which is noisy on the logs). I might try to contribute that API, but I didn't discuss it with upstream yet. Probably independent on this PR.
- BTW independent of this proposal (I noticed it when changing the sleep code) - the sleep of scan_delay always completes to full even if the other thread is just half a second too slow to signal thread stop - this introduces a long delay until actual decoding begins. Is this intentional?

### DIFF
--- a/auto_rx/autorx/scan.py
+++ b/auto_rx/autorx/scan.py
@@ -865,11 +865,37 @@ class SondeScanner(object):
                 # Scan completed successfuly! Reset the error counter.
                 self.error_retries = 0
 
-            # Sleep before starting the next scan.
-            time.sleep(self.scan_delay)
+            if self.sonde_scanner_running == False:
+                self.log_info("Stopping thread just before sleep")
+                break
+            if _results:
+                # Sleep before starting the next scan.
+                # time.sleep(self.scan_delay)
+                for i in range(self.scan_delay):
+                    time.sleep(1)
+                    if self.sonde_scanner_running == False:
+                        self.log_info("Stopping thread during sleep!")
+                        break
+            else:
+                self.run_readsb(self.scan_delay)
+                if self.sonde_scanner_running == False:
+                    self.log_info("Request to stop thread came during readsb")
 
         self.log_info("Scanner Thread Closed.")
         self.sonde_scanner_running = False
+
+    def run_readsb(self, timeout=60):
+        """Run readsb for a given time period to interleave sonde scans with ADSB traffic scanning.
+
+        XXX: This is all hardcoded and just punk.
+
+        If you run readsb and tar1090 using systemd, you must first do `sudo ln -s /tmp/readsb-run/ /run/readsb; mkdir -p /tmp/readsb-run` on system start."""
+        self.log_info(f"Running readsb for {timeout}s")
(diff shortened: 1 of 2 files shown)


darksidelemm:
I don't really want to deal with any kind of handling of other software within auto_rx. RTLSDRs are cheap enough now that the simplest solution is to just buy another one rather than trying to time-share. Part of the point of auto_rx was to have something watching the band continuously to look for launches which might not be happening at normal times. 
Trying to do everything with a single RTLSDR and antenna is just going to result in a compromised system. 

> BTW independent of this proposal (I noticed it when changing the sleep code) - the sleep of scan_delay always completes to full even if the other thread is just half a second too slow to signal thread stop - this introduces a long delay until actual decoding begins. Is this intentional?

This one i haven't looked too much into, though I think I know the behaviour you mean. I'm guessing some tweaking of those delays would help avoid this?

pasky:
Sorry for the delayed reply.

In the end it's of course your decision, just to explain my motivation better:
- For me the reason to use a single RTL-SDR is that my RPi3B cannot handle decoding two and RPi4 is impossible (or absurdly expensive) to obtain here right now. (And two RPi don't fit into the roof box...) So unfortunately it may not be as simple as "just get a second RTL-SDR".
- The timesharing seems like a fine compromise in practice as this will catch even unscheduled launches, we still check every minute or so (I actually use scan delay 120s but that still seems like an insignificant time). I don't think that's a terribly meaningful loss of telemetry data?
- The footprint of the change is actually super small, and possibly could be used for other purposes if there is a "generic command to run during the downtime" (I can't come up with examples though, admittedly). Would it interfere with some future plans you'd have for the functionality?

> > BTW independent of this proposal (I noticed it when changing the sleep code) - the sleep of scan_delay always completes to full even if the other thread is just half a second too slow to signal thread stop - this introduces a long delay until actual decoding begins. Is this intentional?
> This one i haven't looked too much into, though I think I know the behaviour you mean. I'm guessing some tweaking of those delays would help avoid this?

I dealt with it like this (the log message usually comes about one second into the scan delay).

```
-            # Sleep before starting the next scan.
-           time.sleep(self.scan_delay)
...
+                for i in range(self.scan_delay):
+                    time.sleep(1)
+                    if self.sonde_scanner_running == False:
+                        self.log_info("Stopping thread during sleep!")
+                        break
```

darksidelemm:
Odd that a 3B doesn't handle 2 RTLSDRs - I've run that configuration up at my prime receive site for a while now, before recently switching to an Airspy Mini configuration.

I'll fix the scan delay issue, but I don't really want to add in any other timesharing code within auto_rx, sorry...

PR #783, opened 2023-07-03
Enabling MacOS

mastmees:
Here are the code changes I made for running auto_rx on MacOS. The main things are choosing appropriate existing timeout command and ensuring the main program terminates even if worker threads are not joining. The failure to join is mostly caused by external utilities hanging, which in turn is caused by instability of SDR hardware/drivers/utilities.

The main program now returns with consistent exit codes, so that external SDR resetting script can be used to perform hardware reset on non-linux platforms.

With these changes, auto_rx.sh has been running unattended on my mac for many days without issue.  Feel free to use what you see fit for the project.

Cheers.


darksidelemm:
I develop and test auto_rx on OSX and a Macbook Pro M1, though I don't usually run it for very long periods.

There's a fair few changes in here that I need to look at. A few points:
- Can you please change this PR to target the testing branch.
- For timeout, i just installed gnu coreutils via macports / homebrew. This provided 'timeout' for me.
- Your SDR reset solution isn't really suitable for inclusion into this repository. I'm curious whether you've seen the same kind of RTLSDR lockups on a Mac as we've seen on other platforms.

mastmees:
new PR created against testing branch

PR #784, opened 2023-07-04
Enabling MacOS

mastmees:
creating PR against testing branch as asked. Additionally, found one more place that prevented clean shutdown when SDR locks up, and added schematics for power-cycling hardware.


darksidelemm:
Turns out this exception-based approach actually causes other issues, as sometimes the rtlsdr reset happens within a thread, so the main auto_rx.py process never sees the exceptions. I'm probably going to have to revert a bunch of these changes to avoid auto_rx locking up.

darksidelemm:
OK, so the particular issue was occuring if an sdr fault occured on the SDR that was being used for scanning. 
Since the check and attempt to reset occurs within the scan loop thread, the main process never sees these exceptions
I've add in code which handles the exceptions. What'll happen is it'll try a few times to reset, the scan loop will eventually exit, and then when it tries to start the scan loop again it'll run the sdr test which will trigger the exception that the main process *does* see...

Bit of a long-winded way of handling it, but I don't see much other option.

PR #789, opened 2023-07-24
Fix reverse proxy path handling

### LINKED ISSUE #778: Web UI not usable when behind a reverse proxy
All references to resources in the web UI are absolute (e.g. /static/js/jquery-3.6.0.min.js). If you configure a reverse proxy to protect access to the  web server you are forced to use (a different port and) a path of /, otherwise all references to resources will be broken.
I think it should be quite easy to just drop the leading / to permit use of an arbitrary path in the reverse proxy

drid:
Fixes #778

### DIFF
--- a/auto_rx/autorx/static/js/autorxapi.js
+++ b/auto_rx/autorx/static/js/autorxapi.js
@@ -2,7 +2,7 @@
 
 function update_task_list(){
     // Grab the latest task list.
-    $.getJSON("/get_task_list", function(data){
+    $.getJSON("get_task_list", function(data){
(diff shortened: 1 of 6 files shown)


darksidelemm:
Please re-point this pull request to the testing branch, where such things are included and tested carefully!

buttim:
Sure. I just have to figure out how I can get your commit onto my raspberry. Will do it tomorrow I hope

darksidelemm:
OK I was able to re-point this PR to the testing branch. I need some time to try it out locally before I merge it in.

PR #791, opened 2023-08-01
Use fixed-point notation for ascent rate

argilo:
In the KML feed, Ascent Rate is currently displayed in scientific notation, which is difficult to read:

Screenshot from 2023-08-01 09-30-09

Here I've changed it to fixed-point to match the other fields:

Screenshot from 2023-08-01 09-49-48


darksidelemm:
Thanks for that, I wonder how I missed the 'f' at the end of that format string...

PR #809, opened 2023-09-15
Update email_notification.py

thetoivonen:
Change initial altitude in the landing notification test to be compatible with the new requirement that the sonde exceeds 1000' before being eligible for a landing notification.


darksidelemm:
This has been resolved in the testing branch.

PR #823, opened 2023-10-03
Mainboard type on web interface, upload RS41 subframe data to SondeHub

darksidelemm:
... and optionally save it to disk.


darksidelemm:
testing this one locally for a few days before merging it in...

PR #856, opened 2024-01-27
Write rtl_power log output to user-defined log directory

darksidelemm:


argilo:
This change has caused the following error to be printed when launching with `auto_rx.sh`:

```
rm: cannot remove 'log_power*.csv': No such file or directory
```

Perhaps the `rm log_power*.csv` is not needed in that script?

darksidelemm:
I never use the script, but yes that line can be removed. I've done that in the testing branch.

PR #857, opened 2024-01-28
Combine export_selected and export_all

argilo:
While looking at web.py, I noticed that `flask_export_selected_log_files` and `flask_export_all_log_files` are almost identical. By combining them, some code duplication can be avoided.


darksidelemm:
Can you please re-target this to the testing branch?

argilo:
Will do.

argilo:
Done.

darksidelemm:
I suspect I could have probably just done that from the Github interface, but thanks anyway :-)

PR #858, opened 2024-01-29
Generate KML files from the "Historical" page

argilo:
Log files can be converted to KML using the `log_to_kml` command-line utility. To make the process simpler, I've added a "Generate KML" button to the "Historical" page:

Screenshot from 2024-01-29 17-13-23

When clicked, a KML file is generated from the selected log files.

To avoid making `fastkml` and `shapely` dependencies of auto_rx, I rewrote `log_to_kml` to generate KML using the standard library's built-in XML support. (It should be possible to eliminate auto_rx's `simplekml` dependency in the same way, but I'll leave that to another pull request.)

### DIFF
--- a/auto_rx/autorx/log_files.py
+++ b/auto_rx/autorx/log_files.py
@@ -14,6 +14,7 @@ import logging
 import os.path
 import time
 import zipfile
+import xml.etree.ElementTree as ET
 
 import numpy as np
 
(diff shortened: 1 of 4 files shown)


darksidelemm:
Nice addition!
I never really intended for code in the utils directory to be called from within auto_rx, so it might be best to move the required functions from there to somewhere in the autorx 'module'. The log_files.py file might be a suitable spot.

argilo:
That sounds like a good idea. I'll take a crack at it.

argilo:
OK, KML generation has now been moved into `autorx.log_files`.

In order to use that from `utils/log_to_kml.py`, I had to use a hack to allow an import from the parent directory. Is that OK?

There are a couple leftover `print()` statements in `log_files_to_kml` that came from the `log_to_kml` script. Any opinion on whether we should get rid of those? The bare `except` is also a bit of a pain, since it swallows up all errors and hides them. I might take a look to see what errors can actually occur in practice. Skipping missing or corrupt log files is probably useful.

darksidelemm:
I think most of the print statements should go to logging.debug calls, except those within except blocks. Those should be logging.exception, and should include the exception info to assist with debugging. We should just skip over those broken files though.

argilo:
I replaced the `print` statements with `logging`.

darksidelemm:
OK, are you happy enough with this for me to merge it in?

argilo:
Yes, I think so. It's working well for me, and I verified that exceptions are logged when they occur.

darksidelemm:
Awesome, I'll merge it in!

argilo:
Thanks!

Are you happy with how the `ElementTree` KML generation turned out? If so, I might take a crack at replacing `simplekml` with that, which would get rid of a dependency.

darksidelemm:
Yeah, it seems to work, so I'd be happy to see the dependency removed!

Ran a few tests on my local station with the new code, seems to work fine! As expected exporting everything takes quite a while. I suspect that at some point the URL for large selections is going to get too big (e.g. if you select all, then de-select one), but oh well.

darksidelemm:
Actually it might not be an issue (seems most browsers have stupidly large URL limits), so happy to ignore it for now :-)

PR #859, opened 2024-01-31
Generate live KML using ElementTree

argilo:
Here I've switched live KML generation over to Python's built-in `ElementTree`, which allows it to share a lot of code with the `log_to_kml` utility and eliminates the dependency on `simplekml`. I also modified the output of `log_to_kml` a bit to make it more like live KML, which I think makes it a bit nicer to work with.


argilo:
I added one more small improvement: sondes are now sorted from newest to oldest.

PR #860, opened 2024-01-31
Remove unused Habitat code

argilo:
While working on auto_rx I've been bumping into code related to Habitat. As far as I can tell, this code is obsolete and can be safely removed, which I've done here. In addition to reducing the maintenance burden, this also eliminates a dependency (crcmod).


darksidelemm:
Awesome job with this! 
Yes, this code is all obsolete. Thankfully we left the Habitat DB behind a long time ago...

PR #867, opened 2024-03-27
Replace deprecated utcnow() with now()

argilo:
As of Python 3.12, [`datetime.utcnow()` is deprecated](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow), with the recommended replacement being `datetime.now(timezone.utc)`. I made that replacement here. This was a simple search-and-replace, except in `listener_nmea_crlf.py` where I modified the code to be more like the other scripts. So far I haven't done much testing apart from firing up auto_rx.

### DIFF
--- a/auto_rx/autorx/scan.py
+++ b/auto_rx/autorx/scan.py
@@ -969,7 +969,7 @@ class SondeScanner(object):
             (_freq_decimate, _power_decimate) = peak_decimation(freq / 1e6, power, 10)
             scan_result["freq"] = list(_freq_decimate)
             scan_result["power"] = list(_power_decimate)
-            scan_result["timestamp"] = datetime.datetime.utcnow().isoformat()
+            scan_result["timestamp"] = datetime.datetime.now(datetime.timezone.utc).isoformat()
             scan_result["peak_freq"] = []
             scan_result["peak_lvl"] = []
 
--- a/auto_rx/utils/plot_sonde_log.py
+++ b/auto_rx/utils/plot_sonde_log.py
@@ -417,7 +417,7 @@ def process_directory(log_dir, output_dir, status_file, time_limit = 60):
 
 
             # Calculate the age of the last data point in minutes.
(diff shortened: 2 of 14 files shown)


auto_rx/utils/plot_sonde_log.py
```diff
@@ -417,7 +417,7 @@ def process_directory(log_dir, output_dir, status_file, time_limit = 60):
 
 
             # Calculate the age of the last data point in minutes.
-            _data_age = (pytz.utc.localize(datetime.datetime.utcnow()) - parse(last_time)).total_seconds() / 60.0
+            _data_age = (pytz.utc.localize(datetime.datetime.now(datetime.timezone.utc)) - parse(last_time)).total_seconds() / 60.0
```
darksidelemm, line 420: This will need the pytz.utc.localize( ) call removed, as that expects a naive datetime object.

argilo: Good catch. Fixed.

auto_rx/autorx/scan.py
```diff
@@ -969,7 +969,7 @@ def sonde_search(self, first_only=False):
             (_freq_decimate, _power_decimate) = peak_decimation(freq / 1e6, power, 10)
             scan_result["freq"] = list(_freq_decimate)
             scan_result["power"] = list(_power_decimate)
-            scan_result["timestamp"] = datetime.datetime.utcnow().isoformat()
+            scan_result["timestamp"] = datetime.datetime.now(datetime.timezone.utc).isoformat()
```
argilo, line 972: I noticed that this is going to change behaviour:
```
>>> datetime.datetime.utcnow().isoformat()
'2024-03-27T22:36:59.792134'
>>> datetime.datetime.now(datetime.timezone.utc).isoformat()
'2024-03-27T22:37:09.889948+00:00'
```
Maybe it should get a more specific time format?

PR #877, opened 2024-05-24
auto_rx/build.sh: exit immediatelly on error

pinkavaj:
This ensures script will not continue to build when error accures. Continuing the build after some part fails only brings more errors and makes the output harder to understand.


darksidelemm:
Thanks!

PR #878, opened 2024-05-24
fix error: type defaults to ‘int’

pinkavaj:
This code used STD C99 feature, I kind of expect only modern enough compilers are used for this project.

### DIFF
--- a/demod/mod/lms6Xmod.c
+++ b/demod/mod/lms6Xmod.c
@@ -25,6 +25,7 @@
  *  author: zilog80
  */
 
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


darksidelemm:
Just checking with @rs1729 that this change will be OK?

rs1729:
`lms6mod.c` is deprecated, only there for historical reasons. It was integrated in `lms6Xmod.c` with
`static int vitCodes_init = 0;` (no missing int).
`lms6Xmod` can decode another variant of LMS-6, don't know if it was only a test version though.

Regarding `bool`,
sorry, no, never used bool, don't see the advantage here. And I would have to go through all the other decoders.
I don't want to use "true", for me `0==false`, so it is either 0 or not 0.
It is possible that I'm using degrees of "true" somewhere.

I guess `lms6mod.c` can be removed in auto_rx.

PR #879, opened 2024-05-25
delete lms6mod - replaced by lms6Xmod

pinkavaj:
I'm getting the error mentioned bellow during build, deprecated, so mey-be this is the way?

```
cc -O3 -w -Wno-unused-variable -DVER_JSN_STR=\"1.7.3-beta11\"   -c -o lms6mod.o lms6mod.c
lms6mod.c:180:8: error: type defaults to ‘int’ in declaration of ‘vitCodes_init’ [-Wimplicit-int]
  180 | static vitCodes_init = 0;
      |        ^~~~~~~~~~~~~
make[1]: *** [<builtin>: lms6mod.o] Error 1
```


rs1729:
Yes, I believe it can be deleted in the test tool. In the decoding chain only lms6Xmod is in use.
But @darksidelemm  knows auto_rx better.

darksidelemm:
The test_demod tool has a lot of old stuff within it, I'd just leave it alone - it's not really part of the main codebase. 
But yes, removing lms6mod (and probably a few other old demodulator files) is fine.

PR #914, opened 2024-10-02
Put X axis ticks at exact MHz values

argilo:
The X axis of the scan plot is unwieldy, with hundreds of tick marks and odd frequencies displayed:

Screenshot from 2024-10-02 17-31-24

Providing a fixed list of frequencies cleans things up nicely:

Screenshot from 2024-10-02 17-29-25

I included frequencies for both the 403 MHz and 1680 MHz bands. For the 1680 MHz band, I chose a 2 MHz spacing to avoid having too many labels when the full 1675-1700 MHz band is scanned.

### DIFF
--- a/auto_rx/autorx/static/js/scan_chart.js
+++ b/auto_rx/autorx/static/js/scan_chart.js
@@ -53,10 +53,13 @@ function setup_scan_chart(){
 	    axis:{
 	        x:{
 	            tick:{
-                    culling: {
-                        max: window.innerWidth > 1100 ? 10 : 4
-                    },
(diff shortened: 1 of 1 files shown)


darksidelemm:
This is certainly an improvement! I think i'd prefer 0.5 MHz steps for the 400 MHz band though - that should be easy enough?

argilo:
Yes, that's a simple change. Done.

Screenshot from 2024-10-02 18-24-21

argilo:
It looks like there's enough room for the U.S. 400-406 band even on mobile:

Screenshot_20241002-182307

PR #917, opened 2024-10-12
Remove transition from left margin of map

argilo:
When the Historical page is opened, the map is not centered:

Screenshot from 2024-10-11 22-21-15

Closing the Sonde List sidebar also leaves the map uncentered, as does opening or closing the Log sidebar on the main page.

This happens because `mymap.invalidateSize()` is called immediately, but the left margin of the map moves for the next half second due to a CSS transition. The right margin does not have such a transition, and therefore doesn't suffer the same problem.

Removing the CSS transition solves the problem, and the map is then correctly centered in all cases.

### DIFF
--- a/auto_rx/autorx/static/css/main.css
+++ b/auto_rx/autorx/static/css/main.css
@@ -204,7 +204,6 @@ button[disabled]{
 }
 
 #main {
-  transition: margin-left .5s;
   padding: 16px;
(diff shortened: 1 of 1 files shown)


darksidelemm:
Aha! Great find!! I was wondering why this was happening, thanks for the fix!!

PR #921, opened 2024-10-13
Upgrade to Tabulator 6.3.0

### LINKED ISSUE #915: Telemetry table & map displayed incorrectly in Firefox
Upon loading Auto-RX in Firefox, the telemetry table is displayed with excessively wide columns, and the map is not centered:

Screenshot from 2024-10-02 20-59-42

Clicking a column heading restores the columns to their normal sizes, and resizing the browser window restores the map centering.

After some investigation, I traced the root cause to this line of code, which swaps Tabulator's dark theme for a light one:

https://github.com/projecthorus/radiosonde_auto_rx/blob/757925effaadf876152c1d42022e03b129c44594/auto_rx/autorx/templates/index.html#L70

If this line is commented out, then the problems go away (but the telemetry table always displays in dark mode).

If dark mode is selected in the settings pane, then the problems also go away and the page displays correctly upon initial load:

Screenshot from 2024-10-02 21-00-00

I'm not yet sure what an appropriate fix would be.

argilo:
Fixes #915.

Updating Tabulator fixes the issue with the telemetry table (and map) displaying incorrectly in Firefox. I chose the latest available release, 6.3.0. A number of API changes have been made, so I updated the Auto-RX code where needed.

I've tested the Historical page, but haven't yet had a chance to verify that the telemetry table works correctly during a flight. (I'll have a chance in an hour or two.)


argilo:
It's probably easiest to review the code changes while ignoring differences in white space: https://github.com/projecthorus/radiosonde_auto_rx/pull/921/files?w=1

argilo:
I'm tracking a flight now, and the telemetry table looks to be working as expected.

argilo:
I noticed that the source map for Tabulator (which I had added) was rather large, so I took it out to avoid bloating Auto-RX. If anyone needs it for debugging purposes, they can always download their own copy.

argilo:
I also noticed that I forgot to update `jquery_wrapper.min.js`, which is a part of Tabulator. But it turns out Auto-RX doesn't use it, so I'll remove it instead.

PR #926, opened 2024-10-26
Add scan functionality to KA9Q radio mode

ke5gdb:


ke5gdb:
I removed the commits that added the configuration hooks for the location of the KA9Q powers utility. The idea was half-baked and we should be ok with using the system paths.

Dockerfile
```diff
@@ -21,7 +21,21 @@ RUN apt-get update && \
...
+    libfftw3-dev \
+    libhackrf-dev \
+    libiniparser-dev \
+    libncurses5-dev \
+    libopus-dev \
+    librtlsdr-dev \
+    portaudio19-dev \
+    libasound2-dev \
+    uuid-dev \
+    rsync && \
```
argilo, line 38: I suspect you could pare this list down to just `libavahi-client-dev libbsd-dev libfftw3-dev libiniparser-dev libopus-dev` if you only build the targets you need below, i.e. `make -f Makefile.linux tune powers pcmcat`. This might speed up the build a bit.

ke5gdb: You are correct! Removing those extra packages and better defining the make command worked as expected.

PR #929, opened 2024-11-06
Ensure that SSRC values are unique

argilo:
Spectrum scanning doesn't work properly with ka9q-radio because Auto-RX chooses the same RTP SSRC number (which ka9q-radio uses as a channel identifier) for both SDR tests and spectrum scanning. The SSRC value must be a unique 32-bit integer.

Here I've changed the SSRC generation to use the frequency in kHz (rounded to the nearest integer) followed by two digits to indicate the channel type (01 for sonde reception, 02 for SDR tests, 03 for spectrum scanning). I hope this will ensure the uniqueness of SSRC values.


argilo:
Sonde frequencies should be limited to 7 digits, so SSRC values should be 9 digits at most, keeping them well below 2^32.

PR #939, opened 2024-11-15
Remove scan plot data decimation.

darksidelemm:
The data sent to the web interface for the scan plot was being decimated 10x. This looked to be a decision made back in 2018, maybe for performance reasons.

After discussion with @argilo , I've decided to remove this decimation and allow the full resolution data to be seen in the scan plot.

Hopefully this doesn't lead to performance issues!


argilo:
Maybe remove the peak_decimation function and imports too? It can easily be restored from git history if needed.

PR #950, opened 2024-12-03
Invalidate map size when switching to new table page

argilo:
When many sondes are displayed in the table, switching between pages can change the height of the table, which subsequently changes the height of the map. In this situation, `invalidateSize()` needs to be called to prevent the map from rendering incorrectly.

### DIFF
--- a/auto_rx/autorx/templates/index.html
+++ b/auto_rx/autorx/templates/index.html
@@ -740,6 +740,11 @@
                 }
                 mymap.invalidateSize();
             }
+
+            table.on("pageLoaded", function(pageno) {
+                mymap.invalidateSize();
+            });
+
             // Invalidate map size to fix problems with elements resizing.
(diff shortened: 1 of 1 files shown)


whallmann:
Hi Clayton,
good to know you dive some deeper in the web code.
On my phone, there is some unusaual beheivor with amdroid 10, and chrome
browser.
The Buttons of the status text of every stick are so big. On other PCs with
bigger screens, they look normal.
I try to add a Hardcopy here
only want you to know this.
regards Wolf df7pn





Clayton Smith ***@***.***> schrieb am Di., 3. Dez. 2024,
22:43:

argilo:
@whallmann Could you open a separate issue for that? I did notice while looking over the code that font sizes in various places are defined relative to the viewport height, which seems to cause trouble. Text is excessively small in landscape mode, for example.

PR #951, opened 2024-12-03
Use coloured icons for task status

argilo:
When using ka9q-radio, the "Tasking" section takes up a lot of space:

Screenshot from 2024-12-03 16-55-06

Here I'm experimenting with a more compact display, with a circle representing each receiver:

* 🔵 = scanning
* 🟢 = receiving
* ⚪ = idle

Screenshot from 2024-12-03 16-55-33

Hovering over a circle shows the details in a tooltip:

Screenshot from 2024-12-03 16-55-56

Or click on the tasking section to display a list of all tasks:

Screenshot from 2024-12-03 16-56-08

### DIFF
--- a/auto_rx/autorx/static/css/autorx.css
+++ b/auto_rx/autorx/static/css/autorx.css
@@ -61,17 +61,10 @@
 .icon-angle-down:before { content: '\f107'; } /* '' */
 .icon-history:before { content: '\f1da'; } /* '' */
 
-
-#task_status {
-	display: flex;
-	flex-wrap: wrap;
-	gap: 2px;
-}
-
 .sdrinfo-element {
-	margin: 0px 4px;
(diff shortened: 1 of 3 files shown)


argilo:
This seems to work well for my ka9q-radio setup. The only thing I don't like is that for single-channel setups (which will be much more common), the receiver status is hidden away even though it wouldn't take much space. Perhaps we could skip the drop-down list and display the status on the right when there's only a single channel.

argilo:
I made a couple improvements, and I'm now happy with how it looks.

When there are three tasks or fewer, the full list is always displayed:

Screenshot from 2024-12-07 10-56-12

When there are more than three tasks, a summary is displayed, and clicking on it reveals the full list:

Screenshot from 2024-12-07 10-56-24

PR #953, opened 2024-12-07
Reduce vertical size of scan plot

argilo:
The scan plot takes up quite a lot of precious vertical space. Here I've made it smaller by removing the "Scan Results" heading, reducing the vertical size from 320 pixels (the default) to 200 pixels, and removing the legend.

Before:

Screenshot from 2024-12-07 18-10-22

After:

Screenshot from 2024-12-07 18-09-25

### DIFF
--- a/auto_rx/autorx/static/js/scan_chart.js
+++ b/auto_rx/autorx/static/js/scan_chart.js
@@ -16,6 +16,9 @@ function setup_scan_chart(){
 	        ['x_spectra',autorx_config.min_freq, autorx_config.max_freq],
 	        ['Spectra',0,0]
 	    ],
+		colors: {
+			Spectra: "#2f77b4"
+		},
 	    type:'line'
 	};
 
@@ -27,6 +30,9 @@ function setup_scan_chart(){
 	        ['x_peaks',0],
 	        ['Peaks',0]
 	    ],
+		colors: {
+			Peaks: "#ff7f0e"
+		},
 	    type:'scatter'
 	};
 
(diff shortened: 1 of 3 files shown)


auto_rx/autorx/static/js/scan_chart.js
```diff
@@ -16,6 +16,9 @@ function setup_scan_chart(){
 	        ['x_spectra',autorx_config.min_freq, autorx_config.max_freq],
 	        ['Spectra',0,0]
 	    ],
+		colors: {
+			Spectra: "#2f77b4"
+		},
```
argilo, line 21: Removing the legend had the side effect of assigning colours to the datasets randomly, so I've explicitly defined them here.

auto_rx/autorx/templates/index.html
```diff
@@ -1272,7 +1272,7 @@
                     document.getElementById("main").style.marginLeft = "350px";
                     document.getElementById("mySidenav").style.borderRadius = "0px 25px 25px 0px";
                     mymap.invalidateSize();
-                    setTimeout(scan_chart_obj.resize,500);
+                    scan_chart_obj.flush();
```
argilo, line 1275: Calling `resize()` here would blow away the size set in the constructor, so I've replaced the call with `flush()`, which forces a redraw. The 500-ms delay seems to have been unnecessary.

argilo:
On Android. Before:

Screenshot_20241207-181048

After:

Screenshot_20241207-180939

argilo:
I noticed that the "Latest Scan" text butts up against the telemetry table. One or the other might need a margin set.

Screenshot from 2024-12-07 18-47-20

argilo:
I added a 0.25em margin above and below the telemetry table, which keeps it separated from tasking section & scan plot.

PR #961, opened 2024-12-19
Move to collapsable web GUI elements

ke5gdb:


argilo:
@ke5gdb One issue I noticed with this is that initially (when local storage is empty) the telemetry and map sections default to closed. They should probably be open by default, as before.

Screenshot from 2024-12-23 11-41-57

ke5gdb:
@argilo agreed. Let me see if I can get it working that way again.

PR #963, opened 2024-12-23
Change default map/table collapsable behavior to open if no cookie exists

ke5gdb:


auto_rx/autorx/templates/index.html
```diff
@@ -311,15 +311,15 @@
             });
 
             // Check if user has preffered map visiblity.
-            if (getCookie('map') == 'true') {
+            if (getCookie('map') == 'true' || getCookie('map') == null) {
```
argilo, line 314: This should work. A simpler alternative may be to reverse the condition and check for `'false'`, as is done for some other settings.

ke5gdb: Good idea. Corrected in the below commit.

argilo (approved):
I tested this on my main station, and the page now displays as expected when local storage is empty.

PR #971, opened 2024-12-31
Use pcmrecord instead of pcmcat

argilo:
As noted in https://github.com/ka9q/ka9q-radio/issues/66#issuecomment-2552692572, @ka9q has merged pcmcat's functionality into pcmrecord, and pcmcat may be removed or converted to a shell script. Here I've switched over to using pcmrecord, and this is working well on my test station.

Since the required pcmrecord functionality is only available in recent commits, the wiki will need to be updated to recommend a newer commit. (I've tested with 08142683dbc398087a5b7d384b1a36bb24b1eca3).

Since a new commit is required, there's no reason to keep the workaround for the powers bug (#941) so I've reverted it here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,11 +61,11 @@ RUN git clone https://github.com/miweber67/spyserver_client.git /root/spyserver_
 # Compile ka9q-radio from source
 RUN git clone https://github.com/ka9q/ka9q-radio.git /root/ka9q-radio && \
   cd /root/ka9q-radio && \
-  git checkout ff7fe26a12909317b26bdb24a0576db87c15baf2 && \
+  git checkout 08142683dbc398087a5b7d384b1a36bb24b1eca3 && \
   make \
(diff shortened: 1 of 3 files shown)


argilo:
ka9q-radio now depends on libogg-dev, so I added that to the Dockerfile to fix the build.

ka9q:
Glad it's working for you. I think I'm pretty much done with pcmrecord.c for a while after my intense work of the past few weeks. I did add the --raw option.

I'd like to encourage users to also try the --exec option. It's very convenient when you want to invoke a separate instance of some consumer program on each SSRC stream in the same multicast group. You don't have to worry about the SSRC; it and other parameters, like the radio frequency, sample rate, channel count and antenna description are available as macros to the command line that executes your program. You just add channels in the radiod config file and everything else happens automagically.

With the --stdout option you either have to specify the SSRC or it will automatically latch onto the first one it sees and ignore any others.

PR #980, opened 2025-02-08
Update to the latest ka9q-radio commit

argilo:
Changes to ka9q-radio's makefile mean that it now requires more dependencies to build the utilities that Auto-RX needs. Here I've added in more or less everything, except for `rsync`, which is only needed by the `install` target.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,10 +62,10 @@ RUN git clone https://github.com/miweber67/spyserver_client.git /root/spyserver_
 # Compile ka9q-radio from source
 RUN git clone https://github.com/ka9q/ka9q-radio.git /root/ka9q-radio && \
   cd /root/ka9q-radio && \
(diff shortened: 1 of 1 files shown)


argilo:
This change today looks like it may eliminate the need to pull in the extra dependencies:

https://github.com/ka9q/ka9q-radio/commit/e410528ed3ea82d5aad3cf313f06a1d9e4a69936

argilo:
Also, https://github.com/ka9q/ka9q-radio/pull/107 was just merged, so we can get rid of the `COPTS` override. I'll try updating this PR to the latest commit.

argilo:
I'm now running commit 854ef7510a125a95312dabc285c1ba8371f675f2 on my two receivers (both native installs).

argilo:
854ef7510a125a95312dabc285c1ba8371f675f2 has been running well for the past day with 1.8.1-beta6.

Issue #834, opened 2023-11-07
Fix for CORS on SSL


pb1dft:
Can someone fix the code to support ssl reverse proxy with CORS support

My Diff on the code

diff --git a/auto_rx/autorx/web.py b/auto_rx/autorx/web.py
index 1455a3c..c288dca 100644
--- a/auto_rx/autorx/web.py
+++ b/auto_rx/autorx/web.py
@@ -27,6 +27,7 @@ from threading import Thread
 import flask
 from flask import request, abort, make_response, send_file
 from flask_socketio import SocketIO
+
 from werkzeug.middleware.proxy_fix import ProxyFix
 import re

@@ -45,7 +46,7 @@ cli.show_server_banner = lambda *x: None

 # Instantiate our Flask app.
 app = flask.Flask(__name__)
-app.wsgi_app = ProxyFix(app.wsgi_app, x_proto=1, x_prefix=1)
+app.wsgi_app = ProxyFix(app.wsgi_app, x_for=2, x_port=2, x_host=2, x_proto=2, x_prefix=2)
 app.config["SECRET_KEY"] = "secret!"
 app.config["TEMPLATES_AUTO_RELOAD"] = True
 app.jinja_env.auto_reload = True
@@ -55,7 +56,7 @@ flask_app_thread = None
 flask_shutdown_key = None

 # SocketIO instance
-socketio = SocketIO(app, async_mode="threading")
+socketio = SocketIO(app, cors_allowed_origins="*", async_mode="threading")

^^ Last line has a * in my config now would be nice to have that read from the settings file.

Regards Terence

73-de-LZ:
Sorry to rise again the theme about the reverse proxy, it should be fixed since radiosonde_auto_rx ver.1.6.2, but i can't get it working, please some advice or working example of config.

PR #998, opened 2025-04-09
fix CORS support for reverse web proxy deployments

kx1t:
This fixes CORS support for use with SSL via reverse web proxies (like `nginx`) (further: "rwp") for container deployments. The fix was tested both via a rwp, as well as directly to ensure it doesn't break any non-webproxy deployments.

Specifically, it adds the Python flask-cors module to the container and applies CORS enhancements to the creation of the app and the socketio.

This fix does NOT fix the issue/complexity of having websockets work via a rwp -- the transport mechanism will fall back to http(s) polling. However, this works fine.

How to test --
1. A radiosonde_auto_rx container using reverse-web-proxy with this fix has been deployed to <https://kx1t.com/trenton-sonde>
2. The rwp deployed at kx1t.com uses `nginx` with the following `location` definition (and radiosonde_auto_rx is deployed at http://10.244.95.84:88/):
```config
location /trenton-sonde/ {
    proxy_pass http://10.244.95.84:88/;
    proxy_http_version 1.1;
    proxy_redirect / /trenton-sonde/;
    proxy_set_header  X-Forwarded-Prefix /trenton-sonde;
}
```
3. I tested the rwp deployment listed in (1) as well as the direct deployment to <http://10.244.95.84:88>; via the rwp, socketio uses polling; using the direct link, socketio uses websockets. In both cases, the application works equally well.

Last -- I additionally tested the enhancement that was suggested in https://github.com/projecthorus/radiosonde_auto_rx/issues/834 to change` app.wsgi_app` in web.py to:
```python
app.wsgi_app = ProxyFix(app.wsgi_app, x_for=2, x_port=2, x_host=2, x_proto=2, x_prefix=2)
```
Unfortunately, that completely breaks socketio connections - the app works fine with the current code around `app.wsgi_app`

### DIFF
--- a/auto_rx/autorx/web.py
+++ b/auto_rx/autorx/web.py
@@ -38,6 +38,7 @@ from queue import Queue
 from threading import Thread
 import flask
 from flask import request, abort, make_response, send_file
+from flask_cors import CORS
 from flask_socketio import SocketIO
 from werkzeug.middleware.proxy_fix import ProxyFix
 
@@ -56,9 +57,10 @@ app.jinja_env.auto_reload = True
 flask_app_thread = None
 # A key that needs to be matched to allow shutdown.
 flask_shutdown_key = None
-
+# allow all CORS:
+cors = CORS(app, resources={r"/*": {"origins": "*"}})
 # SocketIO instance
-socketio = SocketIO(app, async_mode="threading")
+socketio = SocketIO(app, async_mode="threading", cors_allowed_origins="*")
 
 # Global store of telemetry data, which we will add data to and manage.
 # Under each key (which will be the sonde ID), we will have a dictionary containing:
--- a/auto_rx/requirements.txt
+++ b/auto_rx/requirements.txt
@@ -1,6 +1,7 @@
 python-dateutil
 flask
 flask-socketio
+flask-cors
 numpy
 requests
 semver


argilo:
Why is CORS necessary? Shouldn't all the content be loaded from the same origin, from the browser's perspective?

xssfox:
@kx1t can you please provide a har recording for stock radiosonde autorx (eg when cors is failing)

kx1t:
@xssfox note that both my web proxy and the auto_rx instance for my site allow '*' for CORS, so overwriting the CORS header probably doesn't / shouldn't do anything as it's set to accept anything. The instances that I have deployed all include the fix of this PR already.

From what I read online, the CORS permissions should be added in any case -- they won't hurt for non-rwp deployments, and they will ensure that rwp deployments work error-free

argilo:
> From what I read online, the CORS permissions should be added in any case

Why? CORS is only needed in cases where you want to allow another website ("origin") to access content served by yours, and that shouldn't be the case for reverse web proxy deployments where everything is served from the proxy's origin.

xssfox:
> @xssfox note that both my web proxy and the auto_rx instance for my site allow '*' for CORS, so overwriting the CORS header probably doesn't / shouldn't do anything as it's set to accept anything. The instances that I have deployed all include the fix of this PR already.
> 
> From what I read online, the CORS permissions should be added in any case -- they won't hurt for non-rwp deployments, and they will ensure that rwp deployments work error-free

I was overwriting it to simulate no cors header to demonstrate that it works fine without.

PR #1013, opened 2025-06-14
HeyWhatsThat terrain profile integration

penfold42:
In Live view data table, you can cross launch a terrain profile view from your station to the balloon.  
- It's not a live view, but every request passes the latest data.
- Launched in a new tab

In History view can show the terrain profile
- Opens in a subwindow (like the Skew-T plot)
- Can plot terrain profile for first or last received packet

Main Page:
- Moved sondehub and radiosondy to "links" column
- added heywhatsthat terrain cross launch to "links" column
- "links" column can be shown/hidden

Added heywhatsthat config section to station.cfg:
If the entire section is missing, it defaults to station_lat/lon/alt values
[heywhatsthat]
hwt_lat = -34.758
hwt_lon = 141.1001
hwt_alt = 42.0


darksidelemm:
I'd prefer not having a separate configuration settings for this. Please make it just use the station lat/lon/alt values - its up to the user to ensure these are correct.

Issue #1004, opened 2025-04-30
Filter Telemetry in Real-Time


michaelselfwx:
It seems to be the case that almost every DFM-17 sonde that I track occasionally has one frame that has a completely wrong position. Sometime this can be just 1 more degree latitude/longitude or sometimes it is something like 5 degrees. Either way, it is annoying because it messes with my coverage maps, maximum distances (on radiosondy.info), and makes the UI look wonky.

I would love for there to be a way to filter out these data points that are unrealistically far away from the last data point. I would think the best way to do this is to make sure the distance/time is not more than like 500 kph or something along those lines.

DL2MF:
> I agree this would be useful, but its going to take a bit of thought to do it in a way that won't cause other issues.
> 
> Initial thoughts:
> 
...
> 
>     * Something like a running average on lat/lon/alt might work, but even then you'd need special cases to handle cases like loss of signal, where on re-acquisition the position could jump significantly.

Full ack to all topics, just an idea concerning the last one:
- if the time delta is greater than n seconds the position may move (n * speed ) + spread value

I'm using a similar approach in my landing notification bot to avoid continously landing messages when a sonde is a treelanding in the RX coverage area.

darksidelemm:
Released in 1.8.2!

PR #1023, opened 2025-10-15
Filter telemetry in real-time

DB8LE:
Addresses #1004

This PR implements a simple real-time filter that works by calculating the velocity between the most recent frame received from a sonde to the previous one, and checking if it is over a fixed limit to decide wether to discard the packet or not.

The only cases in which this approach fails:
1. The first packet received from a sonde is invalid. In this case, the invalid packet will be kept, and the next packet will be discarded. After that, reception will continue normally. I don't think this is avoidable with any real-time filter design, as there is no way to know if the first received frame is invalid.
2. There are multiple sequential invalid frames. In this case the first invalid frame will be discarded, following invalid frames will be allowed and the first normal frame after the sequence will be discarded. Following my observations DFM17 sondes only transmit a single invalid packet throughout the whole flight, meaning this wouldn't be an issue.

### DIFF
--- a/auto_rx/autorx/decode.py
+++ b/auto_rx/autorx/decode.py
@@ -1937,7 +1937,7 @@ class SondeDecoder(object):
                 return False
 
             # Run telemetry from DFM sondes through real-time filter
-            if self.enable_realtime_filter and (_telemetry["type"].startswith("DFM")):
+            if self.enable_realtime_filter and (_telemetry["model"].startswith("DFM")):
                 # If sonde has already been received, calculate velocity
                 velocity = 0
                 if _telemetry["callsign"] in self.last_positions.keys():


auto_rx/autorx/decode.py
```diff
@@ -1936,14 +1936,14 @@ def handle_decoder_line(self, data):
                 self.decoder_running = False
                 return False
 
-            # Run telemetry through real-time filter
-            if self.enable_realtime_filter:
-                if self.last_position is not None:
+            # Run telemetry from DFM sondes through real-time filter
+            if self.enable_realtime_filter and (_telemetry["type"][:3] == "DFM") and (self.last_position is not None):
```
darksidelemm: Use _telemetry.startswith("DFM") here.

DB8LE: Not _telemetry["type"]?

darksidelemm:
RS92s have related problems, but its not quite the same problem as the DFM. I wouldn't include RS92s in this. They are also pretty uncommon.

DB8LE:
This should be able to handle multiple sondes being received by one decoder. It's certainly not as simple as the other solution, but should work.

darksidelemm:
Apologies, I haven't had any time to test this. I'm just going to merge it into the testing branch and see how things go.

PR #1027, opened 2025-11-18
Attempt to fix possible causes of async hangups w/out logs

cartpauj:
The async cleanup code was using a shared lock that caused all concurrent scans to wait for each other. This may eventually deadlock the scanner after a few cycles.

Ref: https://github.com/projecthorus/radiosonde_auto_rx/pull/1026#issuecomment-3542190707

  Fixed:
  - Removed unnecessary lock from async cleanup (each frequency has unique SSRC anyway)
  - Added subprocess timeouts to prevent hangs
  - Fixed fallback logic so sequential scanning actually runs when async fails
  - Moved cleanup to finally blocks so KA9Q channels always get closed

  Should hopefully fix the issue where scanning would freeze after a while with concurrent workers enabled.

### DIFF
--- a/auto_rx/autorx/ka9q.py
+++ b/auto_rx/autorx/ka9q.py
@@ -49,8 +49,14 @@ def ka9q_setup_channel(
 
     try:
         _output = subprocess.check_output(
-            _cmd, shell=True, stderr=subprocess.STDOUT
+            _cmd, shell=True, stderr=subprocess.STDOUT, timeout=10
         )
+    except subprocess.TimeoutExpired:
+        logging.critical(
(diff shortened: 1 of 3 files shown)


ryanwoolley:
Sure,

KA9Q is running natively on the same device -- a Pi 4B -- as
radiosonde_auto_rx under Docker.

 Scans probably stop somewhere around 11-26-2025 01:04:xx here:
https://gist.github.com/ryanwoolley/15892583377f87dd75a48e18e68b2763

station.cfg is here:
https://gist.github.com/ryanwoolley/05abe7174e61bdab8360a9aa2013cab1

PR #1039, opened 2026-01-20
Add DETACH_KERNEL_DRIVER flag to librtlsdr build

emuehlstein:
Enable automatic kernel driver detachment for RTL-SDR devices by adding -DDETACH_KERNEL_DRIVER=ON to the cmake configuration. This allows librtlsdr to automatically detach the kernel driver when accessing RTL-SDR devices for better interoperability with multiple devices and containers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ RUN pip3 install \
 RUN git clone https://github.com/steve-m/librtlsdr.git /root/librtlsdr && \
   mkdir -p /root/librtlsdr/build && \
   cd /root/librtlsdr/build && \
-  cmake -DCMAKE_INSTALL_PREFIX=/root/target/usr/local -Wno-dev ../ && \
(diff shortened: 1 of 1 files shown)


darksidelemm:
Does detatching the kernel driver actually work from within a docker container? I't be a little surprised if that was possible!

As part of the setup instructions we ask users to blacklist a bunch of the DVB drivers to try and avoid this being a problem.

snh:
I suspect this would only work if running the container in privileged mode, which is generally recommended against. @emuehlstein Have you had a chance to test this in both scenarios? (Privileged vs non-privileged)?

emuehlstein:
Need to do further testing today on a fresh OS install...  I'm realizing I may have neglected to remove the blacklisted modules from my docker host before my previous tests.

darksidelemm:
This change has been made in v1.8.2, thanks!

PR #1062, opened 2026-05-11
Only scan sonde types feature

cartpauj:
## BLOCKED BY: https://github.com/rs1729/RS/pull/61

Adds a new `station.cfg` option that lets us restrict which sonde types `dft_detect` actually scans for.

Right now every scan cycle runs the signal through ~14 type templates (DFM, RS41, RS92, LMS6, IMET5, MK2LMS, M10/M20, MEISEI, RD94RD41, MRZ, MTS01, WXR301, WXRPN9, IMETafsk) before deciding nothing's there. If you live somewhere that only ever sees one or two of those, the other dozen correlations are CPU work for nothing.

```ini
# Empty list (default) = scan all types, same as today.
only_scan_sonde_types = ["DFM", "RS41"]
```

User-facing names get mapped to the internal `rs_hdr` names inside `scan.py` (e.g. "DFM" → "DFM9", "IMET"/"IMET1"/"IMET4" all enable IMETafsk's detection path, M10 and M20 share one entry). `dft_detect` gets a new `--types` CLI flag that takes the comma-separated list of internal names.

I tested it on a `/dev/zero` input — full scan vs DFM9+RS41 only is about 4x faster on my machine. Real-world will be less dramatic since decode time isn't affected, but for stations stuck in scan mode most of the time it's a real win.

`dft_detect` gets a new `--types` flag with case-insensitive matching, whitespace tolerance, and a warning if you ask for a type that was excluded at compile time (NOC34C50/NOIMET1AB). Config plumbing goes through `config.py`, `scan.py`, `scan_async.py`, and `auto_rx.py`, and both `station.cfg.example` files have docs. Default behavior with no flag passed is identical to before.

### DIFF
--- a/auto_rx/auto_rx.py
+++ b/auto_rx/auto_rx.py
@@ -159,6 +159,7 @@ def start_scanner():
             only_scan=config["only_scan"],
             always_scan=config["always_scan"],
             never_scan=config["never_scan"],
+            only_scan_sonde_types=config.get("only_scan_sonde_types", []),
             snr_threshold=config["snr_threshold"],
             min_distance=config["min_distance"],
             quantization=config["quantization"],
--- a/auto_rx/autorx/config.py
+++ b/auto_rx/autorx/config.py
@@ -80,6 +80,7 @@ def read_auto_rx_config(filename, no_sdr_test=False):
         "only_scan": [],
         "never_scan": [],
         "always_scan": [],
+        "only_scan_sonde_types": [],
         "always_decode": [],
         # Location Settings
         "station_lat": 0.0,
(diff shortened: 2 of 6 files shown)


darksidelemm:
I'm very hesitant to include any changes to dft_detect, and honestly i'm not really sure I see what the improvement here is going to be, apart from maybe a little bit less CPU usage when scanning.

dft_detect typically runs faster than realtime anyway, so the delay is on how long you actually feed dft_detect with information - we default to 5 seconds, but this *could* be lowered (some have tried 1 second with success, though I think this is a bit short). 
This setting is here: https://github.com/projecthorus/radiosonde_auto_rx/blob/master/auto_rx/station.cfg.example#L545
This needs to be at least one full transmit cycle of the radiosonde. To capture intermittent transmissions from sondes like RS41s and M20s, that meanst at least 1 second. 

I would perhaps be willing to include something that ignores bogus scan results on the auto_rx side (e.g. I see a lot of bogus iMet detections).

darksidelemm:
To clarify, if there are going to be any changes in dft_detect, then I would want these made in @rs1729 's upstream repository, with their review. I try and avoid making changes in those upstream utilities as much as possible.

cartpauj:
My fault, I have issue a couple PR's upstream https://github.com/rs1729/RS/pulls

I'll keep my PR's here as drafts and will either remove them, or mark them ready for review depending on upstream merge or not.

Hoping to find a few small, but hopefully meaningful gains on CPU usage during scans.

darksidelemm:
I now essentially have the reverse of this (exclude_types) in the testing branch. I figure this is probably the more commonly needed use-case.

PR #1063, opened 2026-05-12
Use FFTW if available

cartpauj:
## Supported by https://github.com/rs1729/RS/pull/60

Companion to https://github.com/rs1729/RS/pull/60 which adds an optional FFTW backend to `dft_detect`. This just teaches our Makefile to auto-detect `libfftw3-dev` via pkg-config and pass `-DHAVE_FFTW3 -lfftw3f` when it's there. Without the lib, build behavior is unchanged.

Don't merge until the rs1729 PR lands and `dft_detect.c` is re-vendored — until then the define is a no-op (just adds a useless link to libfftw3f if it happens to be installed). Once it lands, dft_detect's forward DFT goes through FFTW for roughly a 4x faster correlation, which mostly shows up as lower CPU per scan rather than faster scans (we're sample-rate-bounded anyway).

### DIFF
--- a/scan/Makefile
+++ b/scan/Makefile
@@ -1,5 +1,22 @@
-CFLAGS = -O3 -w -Wno-unused-variable -DNOC34C50 -DNOIMET1AB
-LDLIBS = -lm
+# Auto-detect single-precision FFTW3 (fftw3f). When available, dft_detect is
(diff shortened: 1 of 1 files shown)


darksidelemm:
This will be added as part of https://github.com/projecthorus/radiosonde_auto_rx/pull/1069

PR #1064, opened 2026-05-20
New Modern UI

cartpauj:
# Modern Web UI Implementation
First off — I really hope I'm not stepping on anyone's toes here. I've been using and loving auto_rx for a while, and I happened to have the time and the skills to put some real work into a UI refresh, so I figured I'd contribute it back. If this isn't a direction you want to go, no hard feelings at all.

What this is: a full rewrite of the web UI as a React SPA (TypeScript + Vite + Tailwind). It replaces the old Jinja templates and lives at the same web port. The Python backend is essentially untouched apart from a new `/save_config` endpoint and its companion `config_writer.py` (preserves comments, ordering, and atomic-writes — your hand-edited station.cfg stays exactly the way you left it). Telemetry, scanning, decoding, logging, uploads — all the auto_rx internals are unchanged.

This is still beta. I've been running it on my own station for a bit and it's been solid for me, but I know my one setup isn't every setup.

**Looking for a few early testers** — the built bundle is committed, so all you need to do is check out this branch and restart auto_rx. Run it for a while and let me know what breaks or feels off. I'll be around to fix things and iterate. Specifically looking for folks running KA9Q or SpyServer to get some feedback on how this UI is working for them.

### Existing auto_rx install

**1. Stop it**
```bash
sudo systemctl stop auto_rx     # or Ctrl+C / docker stop, however you run it
```

**2. Switch branches** (your `station.cfg` is not touched)
```bash
cd /path/to/radiosonde_auto_rx
git remote add cartpauj https://github.com/cartpauj/radiosonde_auto_rx.git
git fetch cartpauj
git checkout feat/new-ui
```

**3. Start it again**
```bash
sudo systemctl start auto_rx
```

Open the web UI at the usual address. Done.

**Rollback:** stop auto_rx → `git checkout testing` → start auto_rx.

---

### New device

Follow the normal auto_rx [install guide](https://github.com/projecthorus/radiosonde_auto_rx/wiki) but clone my fork instead:

```bash
git clone https://github.com/cartpauj/radiosonde_auto_rx.git
cd radiosonde_auto_rx
git checkout feat/new-ui
cd auto_rx
./build.sh
cp station.cfg.example station.cfg
# edit station.cfg with your callsign, location, SDR, etc.
./auto_rx.sh
```

Then set it up as a systemd service [per the wiki](https://github.com/projecthorus/radiosonde_auto_rx/wiki#45-automatic--continuous-operation-via-systemd) if you want it to run on boot.

---

**Secure by default:** web config editing is off unless you set `web_config_enabled = True` in `[web]`. Even when enabled... passwords, email, and command paths are never editable from the browser.

---

Thanks for considering it, and for all the work on auto_rx generally. Below are a few screenshots:

#### Live View
image

#### History View
image

#### Stats View
image

#### Settings Page
image

#### Dark Mode
image

#### Mobile Responsive
image


darksidelemm:
As cool as this looks, this is a huge change, and my concern is around how I'm going to be able to maintain this going forward without much (any!) knowledge in typescript development. 

A few points:
- Python 3.9 no longer works. I get that 3.9 is old now, but it would be good to maintain some backwards compatibility for a while.
- Security issues - I had intentionally excised a bunch of fields from the config so they didn't get exposed via /get_config - this is now gone. Data like the web control password, and email credentials are now visible. There are quite a lot of stations with their auto_rx web interfaces exposed to the web.
- Config changes via web interface - While i can see the benefit of this, I'm concerned this exposes users to more security problems. If someone has a weak password, there are settings that could be changed that could result in machine compromise (e.g. changing the rtl_fm path to some other command). Personally I don't think the config should be editable from the web interface at all. The web interface should be 'read only' for auto_rx, with the minimal exceptions of some limited control over decoders.

I do not have the time to properly review this - this is just from running up the branch and poking around a few things that I came to mind immediately.

cartpauj:
Hey Mark (@darksidelemm) — really appreciate you taking the time to look at this even briefly, and the concerns are all fair. Here's where I've landed on each:

**TypeScript maintenance.**
Totally valid. I don't expect you to learn it to merge this. If you take it I'll commit to being the one on the hook for UI-side fixes for the foreseeable future — you can flag issues in my direction and I'll turn them around. There's also a `UI.md` in the PR that documents the layout, build, and conventions for anyone else who ends up poking at it.

**Python 3.9.**
My bad — fixed. I had a `str | None` annotation at module scope which only works in 3.10+. Changed it to `Optional[str]` so 3.9 boots again.

**Credentials in /get_config.**
You were right, that was a serious regression I introduced. Restored the original 9 `global_config.pop()` calls verbatim (same 9 fields, same place, same order). `web_password` and email creds are no longer exposed.

**Config editing from the web.**
Your point about the rtl_fm path being a remote command injection vector is correct and I should have caught it. I went with a tighter approach instead of removing the editor entirely:

1. New `web_config_enabled` flag in `[web]`, defaults to **False**. When off, the Settings tab is hidden in the UI, the page itself refuses to render, and `/save_config` returns 403. Default deployments get nothing.
2. Even when opted in, the writable surface excludes everything risky: `web_password`, all email credentials, sondehub contact email, and the four command paths (`sdr_fm_path`, `sdr_power_path`, `ss_iq_path`, `ss_power_path`). Those are station.cfg-only now. The UI shows a short explainer where the field used to be.

So out of the box this PR ships read-only behaviour matching your preference. Operators who explicitly want it can flip the flag, and even then the dangerous fields are not editable or seen by the UI.

No pressure on the timeline — if it's still not a direction you want to go, totally understandable. Thanks again either way.

F4EWI:
Hi,
@cartpauj is it possible to test your interface ?
73

cartpauj:
@F4EWI yes, are you planning to try it on a fresh device, or run it on your existing device?

In either case, instructions for both are below:
## Existing auto_rx install

**1. Stop it**
```bash
sudo systemctl stop auto_rx     # or Ctrl+C / docker stop, however you run it
```

**2. Switch branches** (your `station.cfg` is not touched)
```bash
cd /path/to/radiosonde_auto_rx
git remote add cartpauj https://github.com/cartpauj/radiosonde_auto_rx.git
git fetch cartpauj
git checkout feat/new-ui
```

**3. Start it again**
```bash
sudo systemctl start auto_rx
```

Open the web UI at the usual address. Done.

**Rollback:** stop auto_rx → `git checkout testing` → start auto_rx.

---

## New device

Follow the normal auto_rx [install guide](https://github.com/projecthorus/radiosonde_auto_rx/wiki) but clone my fork instead:

```bash
git clone https://github.com/cartpauj/radiosonde_auto_rx.git
cd radiosonde_auto_rx
git checkout feat/new-ui
cd auto_rx
./build.sh
cp station.cfg.example station.cfg
# edit station.cfg with your callsign, location, SDR, etc.
./auto_rx.sh
```

Then set it up as a systemd service [per the wiki](https://github.com/projecthorus/radiosonde_auto_rx/wiki#45-automatic--continuous-operation-via-systemd) if you want it to run on boot.

---

**Secure by default:** web config editing is off unless you set `web_config_enabled = True` in `[web]`. Even when enabled... passwords, email, and command paths are never editable from the browser.

cartpauj:
Closing for now, as I do agree this would be hard to maintain without knowing the languages for it.

PR #1069, opened 2026-06-27
v1.9.0-beta Developmental PR

darksidelemm:
Updates in this testing PR:

- Rebase all decoders and dft_detect
- Use combined m10m20mod decoder instead of separate m10mod and m20mod decoders
- Add CF06/HT03 decoder, using IQ input (fsk_demod has issues with the idle behaviour of these sondes)
- Compile dft_detect with fftw3 if it's available, for some CPU usage reductions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,8 @@ auto_rx/mXXmod
 auto_rx/mp3h1mod
 auto_rx/mts01mod
 auto_rx/weathex301d
+auto_rx/m10m20mod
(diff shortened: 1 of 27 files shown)


darksidelemm:
Some checks done so far:
- dft_detect runs and will detect RS41s and M20's fine.
- RS41 and M20 decoding (using the m10m20 decoder) is working.
- dft_detect builds and runs with fftw3

CF-06/GTH support is in, but is entirely untested.

PR #1072, opened 2026-06-29
Update to ka9q-radio powers handler, rebase ka9q-radio

ke5gdb:
This needs more testing, but it appears to be working well on my SDRplay at the house. I'll test at other sites around the area soon. 

NOTE: The scan time should be lowered to 5 seconds with this change. 20 seconds is too long!


darksidelemm:
We can't change the target ka9q-radio version until the spectrum flip issue described in https://github.com/ka9q/ka9q-radio/issues/191#issuecomment-4816043344 is fixed.

ke5gdb:
Redundant with other PR

PR #1073, opened 2026-06-29
Updates to GitHub Actions script

ke5gdb:


ke5gdb:
This one isn't quite ready for review. I just wanted to see if anything blows up. 

I'm also going to see about setting up some caching so we don't start from the ground up on each build.

PR #1082, opened 2026-07-24
container: ka9q-radio 64bit only

snh:


darksidelemm:
I've added these changes into https://github.com/projecthorus/radiosonde_auto_rx/pull/1081  - thanks @snh !!!

PR #1104, opened 2026-09-04
Perform string formatting for log messages sent to SocketIO

argilo:
Some log messages contain formatting placeholders when displayed in the web interface:

Screenshot from 2026-09-04 13-48-07

This is because the unformatted log message is sent to SocketIO. Formatting the message with `self.format(record)` solves this problem:

Screenshot from 2026-09-04 13-50-57


argilo:
By the way, I was able to trigger that particular log message by connecting to the web server using https instead of http.

